PR #6, opened 2021-07-19
Nixos integration

AntonLydike:
I added a `shell.nix` file and some instructions on how to use this on NixOS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ pip3 install .
 ```
 
 ### Running on NixOS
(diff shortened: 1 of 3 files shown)


botherder:
Thank you your contribution. I am sorry to reject it however, as I would prefer to not open up having to cover every OS/distribution in the documentation. I would be happy to add a dedicated page on the Wiki though, to include the nix configuration files and instructions users would need to know about. If you wish to do so, feel free to share a Gist here and I will incorporate it in the Wiki.

PR #11, opened 2021-07-20
Add implicit creation of output folders

febrezo:
A tiny PR with implicit output folder creation to avoid having to create them manually before conducting each analysis.

### DIFF
--- a/mvt/android/cli.py
+++ b/mvt/android/cli.py
@@ -3,6 +3,7 @@
 # See the file 'LICENSE' for usage and copying permissions, or find a copy at
 #   https://github.com/mvt-project/mvt/blob/main/LICENSE
 
+import errno
 import os
 import sys
 import click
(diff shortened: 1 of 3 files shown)


mvt/android/cli.py
```diff
@@ -54,9 +55,12 @@ def download_apks(all_apks, virustotal, koodous, all_checks, output, from_file):
         if from_file:
             download = DownloadAPKs.from_json(from_file)
         else:
-            if not output:
-                log.critical("You need to specify an output folder (with --output, -o) when extracting APKs from a device")
-                sys.exit(-1)
+            try:
```
AntonLydike, line 58: I think you missed the case where output is None, as you removed the `if not output`. (applies to all changed you made here, maybe you can wrap them in a function as well? Just to keep repetition down)

mvt/android/cli.py
```diff
@@ -54,9 +55,12 @@ def download_apks(all_apks, virustotal, koodous, all_checks, output, from_file):
...
+                os.makedirs(output)
+            except OSError as e:
+                if e.errno != errno.EEXIST:
```
AntonLydike, line 61: You should probably re-throw the error again if it's not handled here

AntonLydike (changes requested):
Maybe take a look at these small oversights?

PR #13, opened 2021-07-20
Prompt for password if none is given

pkirkovsky:
If `--password|-p` is specified with no argument, prompt the user to enter a password interactively.


ThatOneRoadie:
Would recommend this pr as well. It resolves issues with special characters in passwords that bash may not process correctly.

pkirkovsky:
Please delete — branches got messed up. I will create a clean PR

PR #16, opened 2021-07-20
Add Dockerfile with Android dependencies solved

febrezo:


README.md
```diff
@@ -10,7 +10,45 @@ Mobile Verification Toolkit (MVT) is a collection of utilities to simplify and a
...
+
+If this is correct, close the container (`exit`) and it is time to connect the Android device to analyse to the USB port using the development mode as explained in the official docs [here](https://developer.android.com/studio/debug/dev-options).
+To have visibility of the USB, the container WILL need to have access to the USB which is not activated in Docker by default.
+This can be done using the `--privileged` parameter when launching Docker as follows and mounting the USB as a volume.
+
+```bash
+docker run -it --privileged -v /dev/bus/usb:/dev/bus/usb mvt
+```
+
+Note that using the `--pivileged` parameter is insecure for a number of reasons explained in detail [here](https://blog.trailofbits.com/2019/07/19/understanding-docker-container-escapes/) as it gives access to the whole system.
```
U039b, line 40: Typo spotted :) `--privileged`

U039b:
I think it would be nice to have both Android and iOS tools packaged in this Docker image. AFAIK, the version of `libimobiledevice` provided by APT repository does not support the latest version of iOS, so it would be really helpful to build the upstream version in this image.

U039b:
The Dockerfile I wrote based on your own seems to work: https://gist.github.com/U039b/3ad88e2286de99a15409f9a869c92c89

Nicceboy:
Hello guys, I packaged this tool for CinCan project as Docker image, see if it provides any help https://gitlab.com/CinCan/tools/-/blob/master/stable/mvt/Dockerfile. Tool readme etc available in [here.](https://gitlab.com/CinCan/tools/-/tree/master/stable/mvt)

It has wrapper script for both Android and iOS tools, but Android backup extractor is missing from the image.

Image has udev rules, so root user is not required and you don't need `--privileged` flag to run image, only `--device` parameter is required for ADB bridge. However, there this chance that not all udev rules for the very latest phones are not included.

U039b:
Thank you!

README.md
febrezo: Oops. Corrected in the new pushed. Thanks for spotting it! ;)

U039b:
@Te-k  this PR looks good to me, you should merge it :) 
The Docker image fixes the compatibility issue of `libimobiledevice` with the latest versions of iOS.

Issue #9, opened 2021-07-19
[Android] Fix path to su binary


hant0508:
Path for `su` binary is hardcoded in `_adb_check_if_root()` (in the file _android/modules/adb/base.py_) as `/sbin/su`. It's not the only possible location of this binary. On my device running Android 11 rooted with Magisk the path is `/system/bin/su`. According [this StackOverflow answer](https://stackoverflow.com/a/55105140) you should check about 40 paths where `su` binary can be placed by different root providers, or find some other way to check the device is rooted.

botherder:
Closing this as resolved.

PR #19, opened 2021-07-20
Better check for if device has root

goshawk22:
"which su" will return the path of the su binary, or it will return nothing. 
The python boolean of a string with something in it (such as the path of the su binary), will be True.
An empty string (where there is no su binary) will be False.

This fixes Issue #9

### DIFF
--- a/mvt/android/modules/adb/base.py
+++ b/mvt/android/modules/adb/base.py
@@ -96,7 +96,7 @@ class AndroidExtraction(MVTModule):
         """Check if we have a `su` binary on the Android device.
         :returns: Boolean indicating whether a `su` binary is present or not
         """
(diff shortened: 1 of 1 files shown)


U039b:
Looks good to me!

mateuszjablonski:
Actually `command -v su` would be better and safer. Explanation can be found here: https://stackoverflow.com/a/677212

goshawk22:
Thanks for the suggestion!

thyarles (commented):
Sounds good.

Trigus42 (approved):
Works on Android 11 with latest Magisk canary build.

advaithm:
can confirm this works without issues

botherder:
Thank you, merged.

PR #21, opened 2021-07-20
Prompt for password if none is given

pkirkovsky:
If `--password|-p` is specified with no argument, prompt the user to enter a password interactively.

(clone of #13)


botherder:

PR #23, opened 2021-07-20
Update README.md

daboxou:
one argument was missing


goshawk22:
It was right before. `pip3 install .` installs mvt from the source in the git repo, whereas `pip3 install mvt` installs mvt from pypi and is documented in the line above.

Te-k:
Indeed, that is a mistake, closing

PR #31, opened 2021-07-20
[iOS CLI] Remove non-existent SYSDIAGNOSE_MODULES reference

recurrence:
SYSDIAGNOSE_MODULES is not present nor is it referenced elsewhere.  I assume this is a dangling reference to something that was removed?

It's causing a NamError exception when running check-iocs via the iOS CLI.


Te-k:
Thanks !

PR #56, opened 2021-07-21
Updating Documentation for Arch Linux

arayofcode:
Added instructions for Arch Linux and updated with a section that can help debug .android folder missing error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@ It has been developed and released by the [Amnesty International Security Lab](h
 
 ## Installation
 
-First you need to install dependencies, on Linux `sudo apt install python3 python3-pip libusb-1.0-0` or on MacOS `brew install python3 libusb`. For Arch Linux, `sudo pacman -S python-pip libusb`. 
+First you need to install dependencies, on Linux `sudo apt install python3 python3-pip libusb-1.0-0` or on MacOS `brew install python3 libusb`.
 
-Then you can install mvt from pypi with `pip install mvt` (use `--user` in case of Arch Linux to avoid conflicts with pacman packages), or directly from sources:
+Then you can install mvt from pypi with `pip --user install mvt`, or directly from sources:
 
 ```bash
 git clone https://github.com/mvt-project/mvt.git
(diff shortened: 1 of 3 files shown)


docs/install.md
```diff
@@ -38,7 +47,15 @@ Then you can install MVT directly from [pypi](https://pypi.org/project/mvt/)
 pip install mvt
 ```
 
-Or from the source code:
+**Note:** In Arch, running the above command can result in python packages and linux packages being stored in same directory which can result in issues. To overcome this, use 
+
+```bash
+pip install --user mvt
```
NicolasCARPi, line 53: I would just recommend the `--user` everywhere, on mac too.

arayofcode: Done

gmes78:
Arch Linux users can use the [AUR package](https://aur.archlinux.org/packages/mvt/) I made, it's probably better to add it to the docs.

As a side note, the warning about pip conflicting with the package manager if you don't use `--user` applies to all distros, not just Arch.

README.md
```diff
@@ -16,9 +16,10 @@ It has been developed and released by the [Amnesty International Security Lab](h
 
 ## Installation
 
-First you need to install dependencies, on Linux `sudo apt install python3 python3-pip libusb-1.0-0` or on MacOS `brew install python3 libusb`.
+First you need to install dependencies, on Linux `sudo apt install python3 python3-pip libusb-1.0-0` or on MacOS `brew install python3 libusb`. For Arch Linux, `sudo pacman -S python-pip libusb`. 
+
+Then you can install mvt from pypi with `pip install mvt` (use `--user` in case of Arch Linux to avoid conflicts with pacman packages), or directly from sources:
```
JulianGro, line 21: `pip` might default to python2 on a lot of systems, so `pip3` or probably better `python3 -m pip` should be used.

JulianGro (changes requested):
Not sure if it even makes sense to provide mouth feeding instructions for a distribution like Arch Linux which is supposed to only be used by people that know what they are doing.

At the very least I really don't the layout of the page work like this. Having one line with "Linux", "MacOS", and "Arch Linux" really doesn't make too much sense. Maybe make a Unix category and put MacOS at the top? Or just make a Linux category.

README.md
arayofcode: That needs to be fixed. Thank you

botherder:
Thank you your contribution. I am sorry to reject it however, as I would prefer to not open up having to cover every OS/distribution in the documentation. I would be happy to add a dedicated page on the Wiki though, if there is specific troubleshooting that Arch Linux users would need to know about.

Issue #65, opened 2021-07-22
List of malicious indicators


wesgladwell:
Hi, So I need a list of malicious indicators to make use of this app? If so, where can I find one.

RoneoOrg:
Not sure, but maybe there: https://github.com/AmnestyTech/investigations

wesgladwell:
> Not sure, but maybe there: https://github.com/AmnestyTech/investigations

Legendary.

pkirkovsky:
See #58

U039b:
You have to use the STIX file available at https://github.com/AmnestyTech/investigations/blob/master/2021-07-18_nso/pegasus.stix2 - it contains the IOCs.

Issue #58, opened 2021-07-21
Check if Pegasus virus infected a device


Vins-222:

pkirkovsky:
Grab the Pegasus IOCs from [here](https://github.com/AmnestyTech/investigations/tree/master/2021-07-18_nso). The important file is `pegasus.stix2`. Then, point the tool to it using the `-i, --iocs` option with `check-backup`

wtroisey:
Thanks for updating doc. Is iTunes backup sufficient to determine if a device has been infected? Or will I need a complete filesystem dump?

PR #85, opened 2021-07-23
Add IOC info to README and main docs

pkirkovsky:
#58 and #65

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ MVT provides two commands `mvt-ios` and `mvt-android` with the following subcomm
 
 Check out [the documentation to see how to use them](https://mvt.readthedocs.io/en/latest/).
 
+For IOCs (indicators of compromise), *including NSO Group Pegasus*, refer to the [Amnesty International investigations repository.](https://github.com/AmnestyTech/investigations)
+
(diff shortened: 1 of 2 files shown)


U039b:
LGTM

stman:
Hello.

Where do we can download those STIX2 files containing the parameters to detect PEGASUS ? It is not specified anywhere in the doc. Thank you.

botherder:
While I do see why in this moment it is important to indicate where to retrieve the STIX2 file for the Pegasus Project investigation, I am conflicted because part of the reason behind choosing STIX2 is to render MVT agnostic to any particular malware. Perhaps it would be useful to have a dedicated page in the documentation where to collect that and other sources of STIX2 definitions, but I would prefer to characterize MVT more explicitly as an agnostic tool that can be used for this and any future investigation.

dkg:
I've just opened #141 to offer a more generic piece of documentation, one that mention's Amnesty's work on Pegasus without elevating it as a singular focus of MVT.  I hope it's useful!

botherder:
Thank you @dkg, I've merged it. Thank you all for this discussion.

PR #102, opened 2021-07-24
Fixes package extraction without uid

TheJackiMonster:
When I tried running `mvt-android check-adb` I got following error during package extraction:
```
14:15:29 ERROR    [mvt.android.modules.adb.packages] Error in running extraction from module Packages: list index out of range
                  Traceback (most recent call last):                                                                          
                    File "/usr/lib/python3.9/site-packages/mvt/common/module.py", line 134, in run_module                     
                      module.run()                                                                                            
                    File "/usr/lib/python3.9/site-packages/mvt/android/modules/adb/packages.py", line 71, in run              
                      uid = fields[2].split(":")[1].strip()                                                                   
                  IndexError: list index out of range
```
So I looked into the code and added a quick fix for it. I assume this is only a problem when a package does not provide a UID. If this should handled differently please let me know.

### DIFF
--- a/mvt/android/modules/adb/packages.py
+++ b/mvt/android/modules/adb/packages.py
@@ -64,6 +64,9 @@ class Packages(AndroidExtraction):
                 continue
 
             fields = line.split()
+            if len(fields) < 3:
+                continue
+
             file_name, package_name = fields[0].split(":")[1].rsplit("=", 1)
             installer = fields[1].split("=")[1].strip()
             if installer == "null":


botherder:
Can you provide some examples of package definitions that trigger this error?

TheJackiMonster:
Here is one example I have encountered:
```
package:/system/app/HwJetPack/HwJetPack.apk=com.huawei.androidx  installer=
```

In similar cases it could also make sense to print warnings about such packages without UID but I'm not fully sure. I just wanted to use the tool in the first place. ^^'

botherder:
Thank you for your contribution, I have changed the code to handle the issue more gracefully. Do let me know if the issue persists.

TheJackiMonster:
But the line causing the issue wasn't about the installer which could not be parsed. It's an indexing error from parsing the uid. So the code still requires an if- or try-block around this line:
```
uid = fields[2].split(":")[1].strip() 
```

botherder:
Ok, should be fixed with:
https://github.com/mvt-project/mvt/commit/2d277d2d143618b42951df03a26043f210f657b9
Although I don't quite understand how the uid field could be missing entirely. I don't have any similar case.

TheJackiMonster:
I also have only this one case I shared. So it's some weird third-party application from Huawei. It's probably not a good sign that they ship an application without uid but maybe it's just the result of broken packaging.

PR #111, opened 2021-07-25
[ADB] Handle SMS base in mmssms.db instead of bugle_db

EmilienCourt:
On my phone, the SMS database is located at `/data/data/com.android.providers.telephony/databases/mmssms.db` instead of `/data/data/com.google.android.apps.messaging/databases/bugle_db`.

The format of the base itself is also different, so the SQL query had to be adapted :

Original SQL query :
```sql
SELECT 
    ppl.normalized_destination AS number,
    p.timestamp AS timestamp,
CASE WHEN m.sender_id IN 
(SELECT _id FROM participants WHERE contact_id=-1)
THEN 2 ELSE 1 END incoming, p.text AS text 
FROM messages m, conversations c, parts p,
        participants ppl, conversation_participants cp
WHERE (m.conversation_id = c._id)
    AND (m._id = p.message_id)
    AND (cp.conversation_id = c._id)
    AND (cp.participant_id = ppl._id);
```

New SQL query added :
```sql
SELECT 
    address AS number,
    date_sent AS timestamp,
    type as incoming,
    body AS text 
FROM sms;
```

To handle both locations, I added a `_adb_check_file_exists` in `mvt/android/modules/adb/base.py` and the `SMS_DB_TYPE` attribute in class `SMS`

### DIFF
--- a/mvt/android/modules/adb/base.py
+++ b/mvt/android/modules/adb/base.py
@@ -112,6 +112,19 @@ class AndroidExtraction(MVTModule):
         :returns: Output of command
         """
         return self._adb_command(f"su -c {command}")
+    
+    def _adb_check_file_exists(self, file):
+        """Verify that a file exists.
+        :param file: Path of the file
+        :returns: Boolean indicating whether the file exists or not
+        """
+
+        # Connect to the device over adb.
+        self._adb_connect()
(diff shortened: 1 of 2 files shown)


botherder:
Thank you.

PR #131, opened 2021-07-29
Chrome History - Cheking extracted URLs against indicators.

macmade:
Indicators are actually not checked for Chrome history records.  
This pull request adds the `check_indicators` method to the `ChromeHistory` class.


botherder:
Thank you.

PR #135, opened 2021-07-30
mvt-ios decrypt-backup: Enable pulling password from the environment.

dkg:
Specifying the password on the command line with `--password XXX`
leaves the password itself visible to any process on the machine which
can scan the process table.

On some systems (including common GNU/Linux distributions) this
visibility is possible by default.

This change should make it possible to offer the password without
putting it into the process table; rather, the user puts the password
in the environment, and specifies the name of the environment
variable, like so:

```
$ export IDEV_BKUP_PASS=WronglySconeRoundnessUnruffled
$ mvt-ios decrypt-backup -d /path/to/dest --passenv IDEV_BKUP_PASS /path/to/data/XXXXXXXX-YYYYYYYYYYYYYYY/
$ unset IDEV_BKUP_PASS
```

I tried first putting the password itself as a UTF-8-encoded string in
a file, and then pointing the `--key-file` option at it, but it looks
like that doesn't follow the same codepath as `--password`.

### DIFF
--- a/docs/ios/backup/check.md
+++ b/docs/ios/backup/check.md
@@ -41,9 +41,10 @@ In case you have an encrypted backup, you will need to decrypt it first. This ca
       -d, --destination TEXT  Path to the folder where to store the decrypted
                               backup  [required]
 
-      -p, --password TEXT     Password to use to decrypt the backup NOTE: This
-                              argument is mutually exclusive with arguments:
-                              [key_file].
+      -p, --password TEXT     Password to use to decrypt the backup (or, set
+                              MVT_IOS_BACKUP_PASSWORD environment variable)
+                              NOTE: This argument is mutually exclusive with
+                              arguments: [key_file].
 
       -k, --key-file PATH     File containing raw encryption key to use to decrypt
                               the backup NOTE: This argument is mutually exclusive
@@ -51,10 +52,10 @@ In case you have an encrypted backup, you will need to decrypt it first. This ca
 
       --help                  Show this message and exit.
 
-You can specify either a password via command-line or pass a key file, and you need to specify a destination path where the decrypted backup will be stored. If `-p` is omitted, MVT will ask for a password. Following is an example usage of `decrypt-backup`:
+You can specify the password in the environment variable `MVT_IOS_BACKUP_PASSWORD`, or via command-line argument, or you can pass a key file.  You need to specify a destination path where the decrypted backup will be stored. If a password cannot be found and no key file is specified, MVT will ask for a password. Following is an example usage of `decrypt-backup` sending the password via an environment variable:
 
 ```bash
-mvt-ios decrypt-backup -p password -d /path/to/decrypted /path/to/backup
+MVT_IOS_BACKUP_PASSWORD="mypassword" mvt-ios decrypt-backup -d /path/to/decrypted /path/to/backup
 ```
 
 ## Run `mvt-ios` on a Backup
--- a/mvt/ios/cli.py
+++ b/mvt/ios/cli.py
@@ -81,9 +81,7 @@ def decrypt_backup(destination, password, key_file, backup_path):
 #==============================================================================
 @cli.command("extract-key", help="Extract decryption key from an iTunes backup")
 @click.option("--password", "-p",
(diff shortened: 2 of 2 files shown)


dkg:
Note that most modern GNU/Linux distributions *do* shield the environment table in one process from any user account other `root` and the user executing the process.  So this only offers a protection against other non-privileged users on the same host (the same set of folks who can't inspect the memory space of the running process).

So this is not a huge security concern (it only really affects multiuser systems, or systems with good privilege separation), but it is a part of a defense in depth strategy.

botherder:
Thank you @dkg for your contribution. I am wondering if from a usability perspective it would be better to impose a pre-determined env variable name (https://click.palletsprojects.com/en/7.x/options/#dynamic-defaults-for-prompts) or to let the user specify one as you suggested. Any thoughts?
FYI, if `--password` is specified without any value it will offer to enter the password from a prompt.

dkg:
> from a usability perspective it would be better to impose a pre-determined env variable name (https://click.palletsprojects.com/en/7.x/options/#dynamic-defaults-for-prompts) or to let the user specify one as you suggested.

I would be fine with either, though the option space you've defined here is fairly overconstrained.  I can propose a patch that:

- uses `--key-file` if provided on the command line
- uses `--password` if provided on the command line
- falls back to a preselected env var like `os.environ['MVT_IOS_BACKUP_PASSWORD']` if neither `--key-file` nor `--password` is present
- if that env var is not set, then it prompts interactively 

> FYI, if `--password` is specified without any value it will offer to enter the password from a prompt

Thanks for that suggestion, i didn't know that, as it's not documented in the `--help` output.

Furthermore, i guess this is only possible to do if you put `--password` at the end of the command line, but the synopsis from `mvt-ios decrypt-backup --help` implies that the options come before the backup path:

```
Usage: mvt-ios decrypt-backup [OPTIONS] BACKUP_PATH
```

I'm clearly reading the synopsis as being more prescriptive than it is, but it does make it a bit confusing.

> (which makes me wonder now if we should enforce the prompt to discourage people from entering the password as a command-line argument).

it's important to me that tools like `mvt-ios` be able to run in a non-interactive mode, so i would be reluctant to "enforce the prompt" without having a clear plan for some other way to get a password-style argument into the program.

botherder:
> I would be fine with either, though the option space you've defined here is fairly overconstrained. I can propose a patch that:
> 
> * uses `--key-file` if provided on the command line
> * uses `--password` if provided on the command line
> * falls back to a preselected env var like `os.environ['MVT_IOS_BACKUP_PASSWORD']` if neither `--key-file` nor `--password` is present
> * if that env var is not set, then it prompts interactively

Yes, this sounds good to me.

> Thanks for that suggestion, i didn't know that, as it's not documented in the `--help` output.
> 
> Furthermore, i guess this is only possible to do if you put `--password` at the end of the command line, but the synopsis from `mvt-ios decrypt-backup --help` implies that the options come before the backup path:
> 
> ```
> Usage: mvt-ios decrypt-backup [OPTIONS] BACKUP_PATH
> ```
> 
> I'm clearly reading the synopsis as being more prescriptive than it is, but it does make it a bit confusing.

You are right, it is generally confusing. It's a little tricky to fully customize things though using the structure that Click provides.

> it's important to me that tools like `mvt-ios` be able to run in a non-interactive mode, so i would be reluctant to "enforce the prompt" without having a clear plan for some other way to get a password-style argument into the program.

Agreed.

pkirkovsky:
> You are right, it is generally confusing. It's a little tricky to fully customize things though using the structure that Click provides.

What are your thoughts on changing BACKUP_PATH to an option instead? It seems passing the path that way allows for mixing the options in any order. I've also considered adding an `--interactive` flag to toggle whether MVT will prompt for things rather than failing, but that is more complicated.

botherder:
I think implementing a failover system as suggested by @dkg  is probably the cleaner way to go.

dkg:
OK, i think i've implemented the designed changes.  I went ahead and applied the same logic to `mvt-ios extract-key` as well so that it can be used in automated fashion without exposing the password to the command line.

botherder:
Looks good, thank you @dkg! Merged.

PR #140, opened 2021-07-31
Avoid breakage with paths with unusual names

dkg:
If file_path has any whitespace or shell metacharacters in it, then
the invocation of subprocess.call would be likely to break (or even
accidentally execute code, depending on how perverse the pathnames
are).

It's generally a good plan to avoid shell=True for subprocess.call
where you can lay out the arguments deliberately in python.  This one
looks relatively straightforward (but note, i have not tested it,
sorry!)

### DIFF
--- a/mvt/ios/modules/fs/base.py
+++ b/mvt/ios/modules/fs/base.py
@@ -50,13 +50,14 @@ class IOSExtraction(MVTModule):
 
         if not shutil.which("sqlite3"):
             raise DatabaseCorruptedError("Unable to recover without sqlite3 binary. Please install sqlite3!")
+        if '"' in file_path:
+            raise DatabaseCorruptedError(f"Database at path '{file_path}' is corrupted. unable to recover because it has a quotation mark (\") in its name.")
 
         bak_path = f"{file_path}.bak"
         shutil.move(file_path, bak_path)
 
-        cmd = f"sqlite3 {bak_path} '.clone {file_path}'"
-        ret = subprocess.call(cmd, shell=True, stdout=subprocess.PIPE,
-                              stderr=subprocess.PIPE)
+        ret = subprocess.call(['sqlite3', bak_path, f'.clone "{file_path}"'],
+                              stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         if ret != 0:
             raise DatabaseCorruptedError("Recovery of database failed")
 


botherder:
I am not sure about this, because without `shell=True` it would fail to me. I would need to do some further tests.

dkg:
Can you give more detail about what kind of failure are you seeing, in what context?  I've tried to simplify this problem to a reproducer here:

[mvt-pr-140.zip](https://github.com/mvt-project/mvt/files/6912903/mvt-pr-140.zip) contains a database (`foo.db`), a test script (`sqlite3-clone.py`), and a copy of the database with a terrible/scary name (`$(touch should-not-exist).db`).

The test script takes the name of the sqlite3 db file to work with.  if the env var `$USE_SHELL` is set to `true`, then it uses the `shell=True` variant of the cloning code. otherwise, it uses the non-shell variant, with the explicit argument list.

here's my attempt to demonstrate that it works with the non-shell variant, and that the shell variant not only fails to clone the database, it silently *fails* (by cloning an empty database), and it executes arbitrary shell commands:

```console
USER@HOST:mvt-pr-140$ ls
'$(touch should-not-exist).db'	 foo.db   sqlite3-clone.py
USER@HOST:mvt-pr-140$ ./sqlite3-clone.py '$(touch should-not-exist).db'
Not using shell
USER@HOST:mvt-pr-140$ sha256sum *
91c9402793e147c79923aa84d27d69e30381fce5415a6ed97dfd53ba4f7bd794  $(touch should-not-exist).db
91c9402793e147c79923aa84d27d69e30381fce5415a6ed97dfd53ba4f7bd794  $(touch should-not-exist).db.bak
91c9402793e147c79923aa84d27d69e30381fce5415a6ed97dfd53ba4f7bd794  foo.db
3948d9b00e6965d51263dcd6a242d5a717010a782369cacb22f14fda37001452  sqlite3-clone.py
USER@HOST:mvt-pr-140$ rm *.bak
rm: remove regular file '$(touch should-not-exist).db.bak'? y
USER@HOST:mvt-pr-140$ USE_SHELL=true ./sqlite3-clone.py '$(touch should-not-exist).db'
Using shell
USER@HOST:mvt-pr-140$ sha256sum *
90ace00a99ea2550a26104bcbe895ed20e9f70b0eac2abc3df509ecf65d989cf  
91c9402793e147c79923aa84d27d69e30381fce5415a6ed97dfd53ba4f7bd794  $(touch should-not-exist).db.bak
91c9402793e147c79923aa84d27d69e30381fce5415a6ed97dfd53ba4f7bd794  foo.db
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  should-not-exist
3948d9b00e6965d51263dcd6a242d5a717010a782369cacb22f14fda37001452  sqlite3-clone.py
USER@HOST:mvt-pr-140$ sqlite3 '$(touch should-not-exist).db' .dump
PRAGMA foreign_keys=OFF;
BEGIN TRANSACTION;
COMMIT;
USER@HOST:mvt-pr-140$ 
```

dkg:
I've also just rebased this, and made a couple more minor changes, as i went chasing down the rabbit hole of potential shell injections, ugh: 

 - i wrapped the sqlite3 `.clone` destination name in quotes (`"`), so that the command works properly when there are spaces in the name.
- to rule out truly malicious injection, i inserted a check that deliberately avoids trying to execute `sqlite3` if the filename contains a `"` character.

botherder:
Thank you @dkg. I used to have issues in getting sqlite3 to accept the .clone command when doing the subprocess.call without shell=True. With your patch I'm not longer able to reproduce it, so either it was an environment error of mine or somehow this has been resolved.

This looks good to me so I will merge this. We'll keep an eye to see if it breaks in other circumstances.
Thanks for the contribution!

PR #141, opened 2021-07-31
Describe how to use and find IOCs

dkg:
This offers generic documentation, to show how MVT can be used with
arbitrary STIX-formatted IOCs, while still pointing users at some
known-to-be-useful sample files.

This aims to be more malware-agnostic than #85.  It might also help search engines recognize the relevance of MVT to people who are already searching for STIX-formatted IOCs.


botherder:
Thank you! I only made some small adjustments and removed the other STIX2 definitions because they were related to Windows malware rather than mobile.

Issue #28, opened 2021-07-20
Decrypt-backup fails despite right password


impartlogic:
Ubuntu 20.02 LTS - installed mvt from source. Enabled backup encryption and set password, took full backup. When I try to decrypt, it fails. If I change the password, it succeeds, so I know it's the right password. 

```
mvt-ios decrypt-backup -p password -d ~/Documents/iphone/00008101-0002*************/ ~/Documents/iphone/output"

``forensics@forensics:~/mvt$ mvt-ios decrypt-backup -p ******** -d ~/Documents/iphone/00008101-************/ ~/Documents/iphone/output
15:31:48 INFO     [mvt.ios.decrypt] Decrypting iOS backup at path              
                  /home/forensics/Documents/iphone/output with password        
         ERROR    [mvt.ios.decrypt] [Errno 2] No such file or directory:       
                  '/home/forensics/Documents/iphone/output/Manifest.plist'     
                  Traceback (most recent call last):                           
                    File "/home/forensics/.local/lib/python3.8/site-packages/mv
                  t/ios/decrypt.py", line 75, in decrypt_with_password         
                      self._backup =                                           
                  iOSbackup(udid=os.path.basename(self.backup_path),           
                    File "/home/forensics/.local/lib/python3.8/site-packages/iO
                  Sbackup/__init__.py", line 172, in __init__                  
                      self.loadManifest()                                      
                    File "/home/forensics/.local/lib/python3.8/site-packages/iO
                  Sbackup/__init__.py", line 1055, in loadManifest             
                      self.date=iOSbackup.convertTime(os.path.getmtime(manifest
                  File), since2001=False)                                      
                    File "/usr/lib/python3.8/genericpath.py", line 55, in      
                  getmtime                                                     
                      return os.stat(filename).st_mtime                        
                  FileNotFoundError: [Errno 2] No such file or directory:      
                  '/home/forensics/Documents/iphone/output/Manifest.plist'     
         CRITICAL [mvt.ios.decrypt] Failed to decrypt backup. Did you provide  
                  the correct password? 

```

Te-k:
`No such file or directory:   '/home/forensics/Documents/iphone/output/Manifest.plist'
It means that you are not providing the right path

Issue #36, opened 2021-07-21
Improve message in mvt-ios decrypt-backup if wrong folder


Te-k:
It is confusing, see #28

botherder:
Closed as resolved.

PR #147, opened 2021-08-02
mvt-ios decrypt-backup: Improve error messages for known cases

dkg:
The two most common reasons that `mvt-ios decrypt-backup` can fail are
wrong passwords and not pointing to an actual backup.

We can distinguish these cases based on the kinds of errors thrown
from iOSbackup (at least for the current versions that i'm testing
with).

When we encounter those particular exceptions, just report a simple
summary and don't overwhelm the user with a backtrace.  If we
encounter an unexpected exception, leave the reporting as-is.

Closes: #28, #36

### DIFF
--- a/mvt/ios/decrypt.py
+++ b/mvt/ios/decrypt.py
@@ -78,8 +78,13 @@ class DecryptBackup:
                                      cleartextpassword=password,
                                      backuproot=os.path.dirname(self.backup_path))
         except Exception as e:
-            log.exception(e)
-            log.critical("Failed to decrypt backup. Did you provide the correct password?")
+            if isinstance(e, KeyError) and len(e.args) > 0 and e.args[0] == b"KEY":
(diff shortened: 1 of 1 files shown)


botherder:
Thank you for this fix. I was planning to do the same. I'm not sure I understand the error message thrown for the FileNotFound error though. What do you mean with "Did you need to specify the full path?".

botherder:
Yes, that's correct. `idevicebackup2` creates backup in the name of the UniqueDeviceID of the device in question (which you can verify launching `ideviceinfo`). I believe it does this to be compatible with iTunes backup folder structures, but I'm not sure.

Your proposed solution sounds good to me!

botherder:
Should I merge this for now or you want to update this pull request?

botherder:
I have merged this for now as I figured a change to automatically seek the right path would probably be best suited in a separate PR. Thanks.

PR #148, opened 2021-08-02
mvt-ios sqlite3 db recovery: fix quoting sent to sqlite3 .clone

dkg:
In b2afce5c79b9305028e1ff7bb226a0ef3d9affc5, the db filename is
wrapped in double-quotes when passing it to the sqlite3 tool's
`.clone` helper command.

For parsing safety, we avoid performing this cleanup if the filename
itself has a double-quote character in it.  Otherwise, a malformed
filename could lead to arbitrary injection into the sqlite3 command.

In be2468004613fe3a40dcf6c8acebd8c866d635d3, the sqlite3 wrapping
changes to single-quotes.  Either the safety check should be amended
to block pathnames with single-quotes, or the sqlite3 wrapping should
revert to double-quotes.

I opted for the latter here because i think single-quotes are more
likely than double-quotes to show up in pathnames (e.g. a folder named
"Daniel's files"), but either change would be fine, of course.

### DIFF
--- a/mvt/ios/modules/fs/base.py
+++ b/mvt/ios/modules/fs/base.py
@@ -56,7 +56,7 @@ class IOSExtraction(MVTModule):
         bak_path = f"{file_path}.bak"
         shutil.move(file_path, bak_path)
 
-        ret = subprocess.call(["sqlite3", bak_path, f".clone '{file_path}'"],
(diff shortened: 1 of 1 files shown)


botherder:
Thank you.

PR #155, opened 2021-08-06
Create dumpsys_notification.py

kshenv:
This module extracts stats on all notifications by apps after last format/factory reset/wipe which are logged by processes. We can check all the apk module which are not cleaned up by any means.


botherder:
Why did you close it?

PR #162, opened 2021-08-11
mvt-ios decrypt-backup: friendlier error handling

dkg:
This series improves on the error handling for `mvt-ios decrypt-backup`, in particular:

- making sure that a failure gets reported as a failure by returning a non-zero
- implementing the friendlier path-finding logic outlined in #147


botherder:
Thank you!

PR #169, opened 2021-08-18
Added availability details to records.md

gregzo:
Added availability details for backup records which require encryption or aren't available anymore in recent iOS versions.

**Only available in encrypted backups:**

* `calls`
* `interaction_c`
* `safari_browser_state`
* `safari_history`

**Only available before iOS 14.7.0:**

* `id_status_cache`


botherder:

PR #170, opened 2021-08-19
Dockerfile lifting

jekil:
Limiting multiple layers to lift docker size (from 883MB to 794MB) and follow best practices (https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers)


botherder:
Thanks!

PR #173, opened 2021-08-21
Use latest Android platform tools

### LINKED ISSUE #154: Install Android Platform Tools from official release in Dockerfile
To avoid using any outdated packages, it would be preferable to change the Dockerfile to install Android Platform Tools from the [official releases](https://developer.android.com/studio/releases/platform-tools) rather than from apt.

arky:
Install android platform tools for linux in Dockerfile. Fixes #154


botherder:

PR #177, opened 2021-08-30
Checking idstatuscache.plist in a dump for iOS>14.7

harsaphes:
The plist file `com.apple.identityservices.idstatuscache.plist` is removed from iPhone backups since iOS 14.7 but the content is still available from a dump in another location.


botherder:
Thank you for sharing this. Is this new location from 14.7 onwards only? I don't recall seeing a file at that location before. If so, is exported in backups? Or how have you managed to extract it?

botherder:
I have confirmed the location of the new file, although I have not yet found a mean to extract it. @harsaphes if you do, please let me know. Merging this, thanks.

PR #185, opened 2021-09-13
fix multi path/file issue

dozenfossil:
For IDStatusCache, LocationdClients and SafariBrowserState modules running on a filesystem, when multiple *_ROOT_PATHS are valid, using `_find_ios_database` will only set as `self.file_path` (and therefore process) the first valid path found.

However, those files can exist at multiple locations, and they should all be parsed.

This can miss IDStatusCache, LocationdClients and SafariBrowserState entries if an empty or incorrect file is at the first indicated path, and valid and additional entries are located at the other paths.

For SafariBrowserState, the `SAFARI_BROWSER_STATE_BACKUP_IDS` was removed as the patch uses `_get_backup_files_from_manifest` instead.

(this fix is similar to what is done for Netusage, SafariFavicon, SafariHistory etc. ; all modules that use `_get_fs_files_from_patterns`)

PS : sorry for the duplicate, messed up

### DIFF
--- a/mvt/ios/modules/mixed/idstatuscache.py
+++ b/mvt/ios/modules/mixed/idstatuscache.py
@@ -51,11 +51,7 @@ class IDStatusCache(IOSExtraction):
                                  result.get("user"))
                 self.detected.append(result)
 
-    def run(self):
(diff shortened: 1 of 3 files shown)


botherder:
Thank you, merged.

PR #186, opened 2021-09-13
fix error for manipulated entries in DataUsage/NetUsage

beneficentboast:
`check_manipulated` will check for missing or manipulated DB entries.

However, if `proc["proc_name"]` is not defined, `len(proc["proc_name"])` in `_find_suspicious_processes` ([here](https://github.com/mvt-project/mvt/blob/e5f2aa3c3d651b1fe8a0401cce82d3eeb6d7e178/mvt/ios/modules/net_base.py#L148)) will throw out an error (as `proc["proc_name"] is None`)

This error will interrupt the whole module, stopping the parsing of the file and not launching `check_manipulated()` or `find_deleted()`.

This patch makes sure this does not happen by checking that `proc["proc_name"]` is not `None`


botherder:
Thank you, merged.

PR #219, opened 2021-12-02
Fix ConfigurationProfiles

workingreact:
This PR will fix some errors that would occur when parsing `configurationprofiles` files.

This will fix two bugs :

- If the plist file is empty, `plist.load` ([here](https://github.com/mvt-project/mvt/blob/356bddc3af79677792f1d7554e4f9f09e1948adf/mvt/ios/modules/backup/configuration_profiles.py#L30)) would fail, which will make the entire module fail
- If the plist contains bytes, they need to be b64 encoded. This was done for `SignerCerts`, but also had to be done for some other fields.


Te-k:
Thanks

Issue #113, opened 2021-07-25
usb1.USBErrorTimeout: LIBUSB_ERROR_TIMEOUT [-7]


marrobHD:
I installed it as [mentioned](https://github.com/mvt-project/mvt#installation) with `pip3 install mvt`.


Logs:
```yaml
root@zorinos:/home/user/yee# adb devices
List of devices attached
xxxxxxxx	device
root@zorinos:/home/user/yee# adb kill-server 
root@zorinos:/home/user/yee# mvt-android check-adb
17:09:31 INFO     [mvt.android.cli] Checking Android through adb bridge                                                                                                                                                                       
         INFO     [mvt.android.modules.adb.chrome_history] Running module ChromeHistory...                                                                                                                                                    
17:09:32 INFO     [mvt.android.modules.adb.chrome_history] Extracted a total of 28140 history items                                                                                                                                           
         INFO     [mvt.android.modules.adb.sms] Running module SMS...                                                                                                                                                                         
         ERROR    [mvt.android.modules.adb.sms] Unable to process file /data/data/com.google.android.apps.messaging/databases/bugle_db: File not found                                                                                        
         INFO     [mvt.android.modules.adb.whatsapp] Running module Whatsapp...                                                                                                                                                               
17:09:45 INFO     [mvt.android.modules.adb.whatsapp] Extracted a total of 1932 WhatsApp messages containing links                                                                                                                             
         INFO     [mvt.android.modules.adb.processes] Running module Processes...                                                                                                                                                             
         INFO     [mvt.android.modules.adb.processes] Extracted records on a total of 1 processes                                                                                                                                             
         INFO     [mvt.android.modules.adb.dumpsys_batterystats] Running module DumpsysBatterystats...                                                                                                                                        
17:09:55 ERROR    [mvt.android.modules.adb.dumpsys_batterystats] Error in running extraction from module DumpsysBatterystats: Could not receive data from first 27d7871f (timeout 10000ms): LIBUSB_ERROR_TIMEOUT [-7]                         
                  Traceback (most recent call last):                                                                                                                                                                                          
                    File "/usr/local/lib/python3.8/dist-packages/adb_shell/transport/usb_transport.py", line 278, in bulk_read                                                                                                                
                      return bytes(self._transport.bulkRead(self._read_endpoint, numbytes, timeout=self._timeout_ms(transport_timeout_s)))                                                                                                    
[...]

PR #229, opened 2021-12-28
Bump adb read timeout

NicolaiSoeborg:
Some adb commands (like `dumpsys`) are very slow and the default timeout is "only" 10s. 
A timeout of 200 seconds is chosen completely at random - works on my phone 🤷

Fixes https://github.com/mvt-project/mvt/issues/113
Fixes https://github.com/mvt-project/mvt/issues/228


botherder:
Thanks.

PR #231, opened 2021-12-30
Search for suid/sgid binaries in common paths

NicolaiSoeborg:
This PR probably requires https://github.com/mvt-project/mvt/pull/229 (if `find` takes more than 10s to run)

More paths could be added.


botherder:
I'm not sure this necessarily fits the intended purpose of the module (which is to highlight binaries specifically related to jailbreaking of the device), perhaps it warrants a new module?

NicolaiSoeborg:
Fair, I will look into making it a new module

PR #241, opened 2022-01-29
Make utf-8 as a default for open()

yallxe:
Not every system uses utf-8 as a default encoding for `open()`, so it is better to write argument `encoding="utf-8"` when using `open()` function in Python. I've tried to find a way to set utf-8 as a default encoding for opening files, but, there is no possibility to do this.

Also, this should fix this errors like this one:
Image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -131,3 +131,6 @@ dmypy.json
 
 # Temporal files
 *~
+
+# IDEA Dev Environment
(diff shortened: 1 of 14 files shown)


yallxe:
Notice: this is usually occure on Windows (WSL also).

botherder:
Thanks.

PR #244, opened 2022-02-01
Update docker.md

dangaffey:
Had to run an additional Docker flag to get it building on the new M1 chip from Apple. Figured it would be helpful to point that out in the Docs for the less initiated users.


goshawk22 (approved):

botherder:
Thanks,

PR #251, opened 2022-02-17
Add initial implementation of SMS extraction using ADB

DonnchaC:
This PR adds support for using the Android backup feature to extract the SMS database directly in MVT. This avoids the need for  either root access or non-standard ADB backup tools such as [ABE](https://github.com/nelenkov/android-backup-extractor).

Still a WIP! Backup encryption is required on many modern Android devices. We will need a basic backup decryption implementation in MVT to decode these encrypted backups.


Te-k:
I made some significant updates:

* Moving the parsing functions to mvt.android.parsers.backup
* Adding support to backup encryption (still no support for compression tho)
* Adding support to mvt-android check-backup

Need discussion still:
* Licence questions
* whether we should move the backup parsing code in the SMS backup module to a higher level class in case we want to have other modules working on backups

Te-k:
It should be good on my side, waiting for review

DonnchaC:
Backup encryption should work with the check-adb SMS module now also. Feel free to merge if it looks good to you.

Te-k:
Looks good to me

PR #259, opened 2022-03-25
base64 encoding fixes in ConfigurationProfiles module

mlowdi:
Certain fields in MDM configuration profile information were not properly base64 encoded when saving output to JSON, causing errors. This is now fixed and tested to work with current Microsoft InTune managed devices.


mlowdi:
Thanks! I'll see if I can figure out how to anonymize a database and share it with you for testing purposes.

PR #266, opened 2022-05-04
add pure-python-adb interface to mvt-android

tothi:
added pure-python-adb support as a replacement for adb-shell in mvt-android.

adb-shell had troubles with tcp-based adb connections (tls is unsupported), while ppadb is not affected by the connection type because it communicates to the local adb server (and the adb server is responsible for communicating with the device).

still not well tested (+no error/exception handling), but seems to work.

### DIFF
--- a/docs/android/adb.md
+++ b/docs/android/adb.md
@@ -37,6 +37,15 @@ mvt-android check-adb --serial 192.168.1.20:5555 --output /path/to/results
 
 Where `192.168.1.20` is the correct IP address of your device.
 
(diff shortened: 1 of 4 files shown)


botherder:
Thanks for the contribution. I'm generally not a fan of adding multiple dependencies. I would prefer to stick to using either one or the other. If pure-python-adb works better, that's great. I picked adb-shell because it looks more actively developed, and @JeffLIrion seems to be responsive to issues being reported. Perhaps if the are missing features we could help contribute those to adb-shell?

I'm open to any suggestions though. Everyone welcomed to chime in.

PR #274, opened 2022-05-25
Adds logging of execution timeline in log file and optional hashing of input files

Te-k:
Hi,

This PR adds two features :
* When an output is given, it logs the execution timeline in `logs.txt` in the output folder
* It generates a hash of the input files in a `hashes.txt` file (optional on large set of data to avoid slowing down the process)

### DIFF
--- a/mvt/android/cli.py
+++ b/mvt/android/cli.py
@@ -22,7 +22,7 @@ from mvt.common.help import (HELP_MSG_FAST, HELP_MSG_IOC,
                              HELP_MSG_OUTPUT, HELP_MSG_SERIAL)
 from mvt.common.indicators import Indicators, download_indicators_files
 from mvt.common.logo import logo
-from mvt.common.module import run_module, save_timeline
+from mvt.common.module import run_module, save_timeline, save_logs
 
 from .download_apks import DownloadAPKs
 from .lookups.koodous import koodous_lookup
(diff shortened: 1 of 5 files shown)


botherder:
I think instead of hashes.txt we should probably have a file including general analysis information/profile, e.g. datetime, path to the analysis target, hashes of the file/s, etc. Probably to go in a dedicated class file, since it's beyond the scope of modules.
We should also refactor CLI commands. The amount of code is piling up in each command and there's a lot of repetition. Thoughts on how to do that nicely?

Te-k:
Good idea for the summary file, what do you think of having it in YAML so it's more redable than json?
I suggest having the following data in it:
* MVT version
* List of IOCs used
* Date
* Path of the data anlysed
* hashes of files
Anything else ?

Re CLI, a large part of the code is the same except the list of modules and the preparsing of the bug report. Could we have a class Analysis that takes a list of modules, a path etc ? And either an inherited class when preparsing is needed, or we pass a function as parameter? Do you see what I mean?

botherder:
Yes, I need to do some thinking and prototyping it.
Regarding the information, sounds good, but I would rather stick to JSON for coherence.

Te-k:
I have updated the code to save an `infos.json` file with information about the analysis and the hashes (optional for backups).

I think this code can be migrated to an analysis class if we want to have one, but for now it can stay in the CLI

PR #279, opened 2022-06-02
Dockerfile improvements, support arm64 builds

Niek:
This PR improves the Dockerfile:
* Switch the base image to LTS 22.04
* Use the native `adb` package (available on arm64 too)
* Docker doc improvements

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:22.04
 
 # Ref. https://github.com/mvt-project/mvt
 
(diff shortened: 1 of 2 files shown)


Niek:
If you want I can also add a workflow that builds the Docker image and pushes it to the GHCR on pushes - let me know if you are interested.

botherder:
Thanks for the contribution.
I don't remember for sure, but I vaguely remember there might have been some issue regarding the version of the package adb potentially being too old?

botherder:
Merging in the meantime.

PR #289, opened 2022-07-02
Fixed cmd_download_apks serial connection bug

lorenzo-reho:
I tried to download apks from my smartphone via serial but it didn't work so I noticed that the Package object was not receiving the string serial, so when _adb_connect() was called from Packages run method the serial variable was set to None and for this reason the connection was not established.


botherder:
Nice find, thanks.

PR #298, opened 2022-08-19
Fixed idevicebackup2 syntax

jons44:
`backup` should be removed from those commands, otherwise it will create a backup in "encryption" or "changepw" directories.


botherder:
Nice find, thanks.

PR #304, opened 2022-09-18
Create ios_GUI.py

ntosis:
initial version, added a small gui with tkinter to simplify the process. 
Only iOS is implemented. 
Auto download from STIX2 files. 
Implement decrypt and check back up steps in one click button. 
No commands in cmd are necessary.  
Auto install from mvt libs via pip if there are not installed.

### DIFF
--- a/gui/ios_GUI.py
+++ b/gui/ios_GUI.py
@@ -1,0 +1,193 @@
+from tkinter import *
+from tkinter.ttk import Treeview,Combobox,Style
+from tkinter import filedialog,messagebox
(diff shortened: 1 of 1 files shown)


DonnchaC:
Hi @ntosis, thank you opening this PR and for your patience with this. We have decided against making a GUI interface for MVT. 

MVT is designed for use by security researchers and forensic investigators. We do not want end-users to have the false impression that MVT is a point-and-click tool to check a device for spyware without additional knowledge to check a device for spyware.

PR #326, opened 2023-01-20
Fixes bug where su binary is present but privilege is not granted to com.android.shell

Hainish:


Hainish:
An exception will be raised otherwise when checking for root permissions.

Te-k:
Thanks !

PR #363, opened 2023-07-14
Add option to disable interactivity

aticu:
This PR adds an option to the `mvt-androidqf` command that disables the interactive password prompt.
In order to pass the argument all the way through, an `interactive` field was added to the `Command` and `Module` classes.

I decided to not change the constructor of `Module`, because then every module that exists would have to change.
Instead the field is set right after the constructor was called.

This was the only instance I could find of interactivity within MVT, but of course if there are other uses of interactivity the flag should be available for their subcommands and apply to them as well.
Did I miss any other uses?

### DIFF
--- a/mvt/android/cli.py
+++ b/mvt/android/cli.py
@@ -4,6 +4,7 @@
 #   https://license.mvt.re/1.1/
 
 import logging
+import os
 
 import click
 
@@ -14,6 +15,8 @@ from mvt.common.help import (
     HELP_MSG_IOC,
     HELP_MSG_LIST_MODULES,
     HELP_MSG_MODULE,
+    HELP_MSG_NONINTERACTIVE,
+    HELP_MSG_ANDROID_BACKUP_PASSWORD,
     HELP_MSG_OUTPUT,
     HELP_MSG_SERIAL,
     HELP_MSG_VERBOSE,
@@ -34,6 +37,8 @@ from .modules.bugreport import BUGREPORT_MODULES
 
 init_logging()
 log = logging.getLogger("mvt")
+
+MVT_ANDROID_BACKUP_PASSWORD = "MVT_ANDROID_BACKUP_PASSWORD"
 CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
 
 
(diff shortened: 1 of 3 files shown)


mvt/android/cli.py
```diff
@@ -277,11 +278,12 @@ def check_backup(ctx, iocs, output, list_modules, verbose, backup_path):
 @click.option("--list-modules", "-l", is_flag=True, help=HELP_MSG_LIST_MODULES)
 @click.option("--module", "-m", help=HELP_MSG_MODULE)
 @click.option("--hashes", "-H", is_flag=True, help=HELP_MSG_HASHES)
+@click.option("--noninteractive", "-n", is_flag=True, help=HELP_MSG_NONINTERACTIVE)
```
DonnchaC, line 281: `--non-interactive` would be better for consistency

mvt/android/cli.py
```diff
@@ -277,11 +278,12 @@ def check_backup(ctx, iocs, output, list_modules, verbose, backup_path):
...
 @click.option("--verbose", "-v", is_flag=True, help=HELP_MSG_VERBOSE)
 @click.argument("ANDROIDQF_PATH", type=click.Path(exists=True))
 @click.pass_context
 def check_androidqf(
-    ctx, iocs, output, list_modules, module, hashes, verbose, androidqf_path
+    ctx, iocs, output, list_modules, module, hashes, noninteractive, verbose, androidqf_path
```
DonnchaC, line 286: `non_interactive` here instead too

mvt/android/modules/androidqf/sms.py
```diff
@@ -58,11 +58,18 @@ def parse_backup(self, data):
...
         password = None
         if header["encryption"] != "none":
-            password = getpass.getpass(prompt="Backup Password: ", stream=None)
+            if self.interactive:
+                password = getpass.getpass(prompt="Backup Password: ", stream=None)
+            else:
+                self.log.warning(
+                    "Cannot decrypt backup, because interactivity"
+                    " was disabled and the password is unknown"
+                )
```
DonnchaC, line 67: Your PR does a good job at avoiding prompting for a password interactively! If this only happens when parsing Android backups then I think we should also add an option to provide a password non-interactively to `mvt-android`.

I imagine a use case where an analyst has collected 50 AndroidQF archives with a common backup encryption password. The analyst would like to run MVT against all of the archives and analyse the enclosed backups too.

MVT should skip prompting for the password if a password is already set as an argument or env variable.

```
mvt-android check-backup --backup-password "ABCD" backup.ab
mvt-android check-backup -p "ABCD" backup.ab
mvt-android check-androidqf --backup-password "ABCD" androidqf_output.zip
mvt-android check-androidqf  -p "ABCD" androidqf_output.zip
MVT_ANDROID_BACKUP_PASSWORD="ABCD" mvt-android check-androidqf androidqf_output.zip
````
`mvt-ios` already has similar functionality from loading a backup password from an argument or an environment variable. 

https://github.com/mvt-project/mvt/blob/main/mvt/ios/cli.py#L99-L126

DonnchaC (changes requested):
Thanks @aticu for sharing this PR. In general it looks good. It would be great if you could also add an option for providing a backup password from an argument or environment variable rather than just skipping it.

If you don't have time then I can also add it later.

aticu:
Thanks for the feedback! I'll try to incorporate these changes.

Also I've found another source of interactivity with the use of `Prompt.ask` in `mvt/android/cmd_check_backup.py`.
In hindsight this is pretty obvious, I don't know why I didn't think to check there earlier.
My only check for other interactivity was grepping for `getpass`, which didn't turn up any other uses.
Is there a specific reason for using `getpass` in one place and `Prompt.ask` in another?

mvt/android/modules/androidqf/sms.py
aticu: I was looking into how to implement something like this and I'm unsure how to pass the backup password down to the `mvt/android/modules/androidqf/sms.py` module.

As far as I could find there's no existing way to pass a module context from the outside (except for the context every module gets of course) or did I miss anything?

The way I would implement this is to give modules access to their parent `Command`, which could then backup password in a field and make it available to the `Module`.
This would be a pretty invasive change, since it would have to change the signature of every module to allow for that change.
What do you think? Is that an acceptable trade-off or do you see a better way of implementing this?

DonnchaC: Good point! We didn't have a good way to pass module specific options down to a command without adding a module parameter which would change the function signature and require changes to every module.

I've just merged a PR #367 which fixed this by creating adding `module_options` parameter. This is a breaking change to any tools which use MVT modules directly but should make things more maintainable going forward, including this PR.

Could you update your PR by passing `non_interactive` and `backup_password` in `module_options` instead? Thanks again for the PR :)

DonnchaC:
> Thanks for the feedback! I'll try to incorporate these changes.
> 
> Also I've found another source of interactivity with the use of `Prompt.ask` in `mvt/android/cmd_check_backup.py`. In hindsight this is pretty obvious, I don't know why I didn't think to check there earlier. My only check for other interactivity was grepping for `getpass`, which didn't turn up any other uses. Is there a specific reason for using `getpass` in one place and `Prompt.ask` in another? Should I maybe unify these to use just one library to make it easier to grep for these uses?

Ah well spotted! No the isn't a reason as far as I know. I suggest using `Prompt.ask` everywhere as that is from the rich library which we are already using extensively for the CLI

aticu:
Alright, I've updated the PR.

The only thing were I'm unsure is the name of the parameter for the backup password.
You've suggested `--backup-password` and I think that's a good name, but an argument could also be made that it should be `--password` for consistency with `mvt-ios decrypt-backup`.
I'm not sure which I prefer, just thought I'd bring it up, before people start depending on it.

DonnchaC (approved):
Great thanks for making those changes! LGTM!

Te-k:
Hi, this has been released in 2.4.1

aticu:
Thank you very much!

tomritzow (commented):
[6.0.md](https://github.com/user-attachments/files/29917506/6.0.md)

tomritzow (commented):
@tomritzow

besendorf:
Please stop spamming our prs

PR #368, opened 2023-07-21
Refactor Android modules to remove duplication

DonnchaC:
Currently we have multiple Android modules extracting the same data from ADB, AndroidQF, and bugreports such as with the `GetProp` modules. Currently each module is a bit different in how it checks for indicators. Some check against Android properties IOCs while others do not. Some also flag interesting properties in the output while others do not.

This WIP PR is a first approach to abstract the detection logic from each of these modules only handler the data extraction process from the respective source. This should be extended to other similar Android modules. We should also consider storing the raw extracted data which can be useful for subsequent analysis.


DonnchaC:
I'm not sure yet is this is the best approach. It may be better to fully abstract the data parsing and detection into separate per artifact modules (ie for GetProp, Dumpsys etc.) as proposed by @Te-k. I'll park this for now until we compare other solutions.

Te-k:
What do you think? One drawback I see is that multiple inheritance is adding complexity, but I think it makes sense conceptually

mvt/android/artifacts/artifact.py
```diff
@@ -0,0 +1,22 @@
...
+        patch_date = datetime.strptime(patch_level, "%Y-%m-%d")
+        if (datetime.now() - patch_date) > timedelta(days=6 * 31):
+            log.warning(
+                "This phone has not received security updates "
+                "for more than six months (last update: %s)",
+                patch_level,
+            )
+            return True
+
+        return False
```
DonnchaC, line 22: I think it would be best to have this function separate to `AndroidArtifact` as I don't think it naturally fits in the generic Android artifact class.

mvt/common/artifact.py
DonnchaC, line 1: It would be helpful to have a short comment about what and how the Artifact class should be used.

This file is also missing a copyright header.

DonnchaC (commented):
Great job on the PR. It looks much cleaner and more maintainable than the duplication in the old implementation. Just some minor comments and then ready to merge!

PR #372, opened 2023-07-23
Adds support for zip files in check-androidqf command

Te-k:
This MR adds support for zip files in `mvt-android check-androidqf`


DonnchaC (approved):
LGTM!

PR #380, opened 2023-08-02
Update applications.py

a-sdi:
Some apps installed from apple store with sourceApp "com.apple.AppStore.ProductPageExtension"


DonnchaC:
Thank you for catching this and contributing!

PR #404, opened 2023-10-02
Fix capitalization of "PyPI"

jwilk:


Te-k:
Is this an hacktoberfest PR?

PR #414, opened 2023-11-13
Add `CustomJSONEncoder` to handle bytes types

roaree:
Adds a custom JSON encoder class to fix serialisation issues where modules included `bytes` types containing non-utf8 bytes, which can't be serialised to JSON.

This adds the `.decode("utf-8", errors="backslashreplace")` option which should replace any invalid UTF-8 bytes with a backslashed escape sequence.


mvt/common/encoding.py
Te-k, line 1: Do you think it deserve a file on its own? Maybe we could have it in utils?

roaree: Sounds good to me! Will move it over

kpcyrd:
How about using either syntax like `"hello\xAAworld"` or `[104, 101, 108, 108, 111, 170, 119, 111, 114, 108, 100]`? That would avoid ambiguity with real string values starting with `0x`.

roaree:
I think `.decode("utf-8", errors="backslashreplace")` is what we're looking for then:

```python
>>> print(json.dumps({"key": b"hello\xaaworld".decode("utf-8", errors="backslashreplace")}))
{"key": "hello\\xaaworld"}
>>>
```

PR #418, opened 2023-11-21
Make `MVTModule.get_slug()` a classmethod

roaree:
Sometimes it can be useful to be able to iterate the module classes and get the "slug"/name before the module is actually initialized. Since modules define their custom `slug` as a class variable, `get_slug()` can be called without an instance of the class.

https://github.com/mvt-project/mvt/blob/73104814ba10127b8c3fd64b4f23cc3dbcc0ccb0/mvt/ios/modules/fs/webkit_indexeddb.py#L16-L25


Te-k (commented):
Sounds good to me!

DonnchaC (approved):
Looks good!

PR #424, opened 2023-11-23
docker tweaks

r-tx:
* Install MVT from github rather than PiPY
* Install latest abe

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,12 +57,12 @@ RUN git clone https://github.com/libimobiledevice/libplist \
 
 # Installing MVT
 # --------------
-RUN pip3 install mvt
+RUN pip3 install git+https://github.com/mvt-project/mvt.git@main
 
 # Installing ABE
 # --------------
 RUN mkdir /opt/abe \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -57,12 +57,12 @@ RUN git clone https://github.com/libimobiledevice/libplist \
 
 # Installing MVT
 # --------------
-RUN pip3 install mvt
+RUN pip3 install git+https://github.com/mvt-project/mvt.git@main
```
roaree, line 60: Ideally I think we should be copying in from whatever's in the locally checked out directory, but this seems like an improvement regardless

roaree (approved):
Thanks!

Dockerfile
r-tx: I could update it for that,  though I wonder would it make more sense for you to put it up on docker hub?  It'd cut down on the steps a little to get the container up and running.

roaree: I would be in favour of us publishing an official docker image

PR #430, opened 2023-11-27
Open all iOS sqlite3 databases with immutable=1

### LINKED ISSUE #429: 2 Filesystem tests fail on macOS only
Only impacts development AFAICT. Would be good to track down a root cause.

```
FAILED tests/ios_fs/test_filesystem.py::TestFilesystem::test_filesystem - AssertionError: assert 17 == 15
FAILED tests/ios_fs/test_filesystem.py::TestFilesystem::test_detection - AssertionError: assert 17 == 15
```

When the tests fail it leaves behind two artifacts on disk:

```
	tests/artifacts/ios_backup/64/64d0019cb3d46bfc8cce545a8ba54b93e7ea9347-shm
	tests/artifacts/ios_backup/64/64d0019cb3d46bfc8cce545a8ba54b93e7ea9347-wal
```

roaree:
Fixes #429 , where `.shm` and `.wal` logs are being left over after tests run.

### DIFF
--- a/mvt/ios/modules/backup/manifest.py
+++ b/mvt/ios/modules/backup/manifest.py
@@ -8,7 +8,6 @@ import io
 import logging
 import os
 import plistlib
-import sqlite3
 from typing import Optional
 
 from mvt.common.module import DatabaseNotFoundError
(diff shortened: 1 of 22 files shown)


roaree:
@DonnchaC / @Te-k I'm not sure if this belongs in `IOSExtraction` or somewhere else?

Te-k:
Looks good to me. If we need it in android modules at some point, we will move it to the MVTModule class.

DonnchaC:
Looks good to me too. Agree that its fine in `IOSExtraction` for now.

PR #438, opened 2023-12-12
Impovements for SMS module

DonnchaC:
The SMS module was creating noisy entries in the timeline as it would output timeline `sms_read` entries for SMS messages even when no `sms_read` timeline was set. This branch fixes the module so that messages which were not read do not produce an extra entry.

This branch also adds basic indicator checking in the SMS module.

### DIFF
--- a/mvt/ios/modules/mixed/sms_attachments.py
+++ b/mvt/ios/modules/mixed/sms_attachments.py
@@ -57,7 +57,6 @@ class SMSAttachments(IOSExtraction):
         for attachment in self.results:
             # Check for known malicious filenames.
             if self.indicators.check_file_path(attachment["filename"]):
-                print("Found malicious filename", attachment["filename"])
                 self.detected.append(attachment)
 
             if (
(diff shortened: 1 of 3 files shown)


Te-k:
LGTM

mvt/ios/modules/mixed/sms_attachments.py
```diff
@@ -55,6 +55,11 @@ def serialize(self, record: dict) -> Union[dict, list]:
 
     def check_indicators(self) -> None:
         for attachment in self.results:
+            # Check for known malicious filenames.
+            if self.indicators.check_file_path(attachment["filename"]):
+                print("Found malicious filename", attachment["filename"])
```
roaree, line 60: Should this use a logger instance?

roaree (approved):
Looks good!

mvt/ios/modules/mixed/sms_attachments.py
DonnchaC: Doh, yes thanks! Not sure how I missed the`print()` when committing.  The `indicators.check_file_path()` already logs the detection. We can just append the attachment to the detections without adding an additional log line.

PR #465, opened 2024-02-20
Add preliminary ipv4-addr ioc matching support under collection domains

renini:
This adds preliminary support to match for any ipv4-addr ioc as if it was a domain.
So any browser activity to http://<ipv4-addr:value> would flag as a known suspicious domain.

Future wise it might make more sense to completely separate ipv4-addr matching logic from the current domain matching logic. But at least for now this would flag as a detection now also for:
https://raw.githubusercontent.com/mvt-project/mvt-indicators/main/2023-07-25_wyrmspy_dragonegg/ip-addresses.txt

### DIFF
--- a/mvt/common/indicators.py
+++ b/mvt/common/indicators.py
@@ -13,6 +13,7 @@ import ahocorasick
 from appdirs import user_data_dir
 
 from .url import URL
+import ipaddress
 
 MVT_DATA_FOLDER = user_data_dir("mvt")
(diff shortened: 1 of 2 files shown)


DonnchaC:
Thanks so much @renini for the PR. We appreciate your contribute and patience on getting it merged. I've now merged #556 which is based on your code. I've removed the CIDR functionality for now and added some additional tests of the functionality.

Thanks again!

PR #476, opened 2024-04-02
Add docs explaining how to seek expert help for forensic analysis 

DonnchaC:
We regularly hear from individuals or groups who have used MVT with public indicators to check their devices for advanced spyware tools. Unfortunately individuals may get a false sense of security if they do not see any matches from the public indicators.  Not all users may fully understand the limitations of MVT to detect recent or advanced attack without expert support and guidance.

This PR updates the documentation to include links to the Amnesty Security Lab website. The PR also adds a log message at the end of each MVT command which explains the limits of MVT and directs non-expert users to seek expert help if they are concerned about spyware targeting.


roaree (approved):

Te-k:
Looks good to me!

PR #488, opened 2024-04-19
Use backwards-compatible datetime.timezone.utc

roaree:
#485 broke timestamp serialisation in python 3.9 only. Due to the broad exception catching, it manifested as "" entries in timelines.

- Removes the broad error catching
- uses `datetime.timezone.utc`, which appears to work for all versions we support


roaree:
Looks like we do actually test on python3.9 here, so will add a regression test also.

PR #489, opened 2024-04-19
Fail builds on test failure

roaree:
As discovered in #488, builds were not failing when tests were failing :(


leitos1980:
Verifcar afectación a mi PC

PR #492, opened 2024-04-26
Fixes datetime issues in iOS SMS parsing

### LINKED ISSUE #466: Error in running extraction from module SMS and SMS attachments
I ran the backup-check on an ios backup and get 2 errors below. The rest of the modules run as expected. What could be the cause of this? I would expect that the SMS table would be available. Any help?

image

image

Te-k:
Since #488, we have to check that dates exists before converting them, which isn't done properly in iOS SMS and SMS Attachments. It should fix #466 but we may have more modules with this issue


roaree:
I agree, we don't know how many modules assume this. I think like you said, it would be better to check for None and return `""`, as before.

Te-k:
Yeah, it may be a better idea, I don't want to chase date issues in all modules, and also it will make the module codes cleaner

PR #527, opened 2024-09-03
Also search for STIX2 files in directories in MVT_STIX2

besendorf:
The old version of this method only accepted paths to files in MVT_STIX2. This new version checks whether a variable in MVT_STIX2 is a file or a dir and then recursively searches for .stix2 files in those dirs.

A change I made is that it only processes .stix2 files which I thought is a reasonably check.


mvt/common/indicators.py
```diff
@@ -40,15 +40,20 @@ def _load_downloaded_indicators(self) -> None:
...
 
         paths = os.environ["MVT_STIX2"].split(":")
         for path in paths:
-            if os.path.isfile(path):
+            if os.path.isfile(path) and path.lower().endswith(".stix2"):
                 self.parse_stix2(path)
+            if os.path.isdir(path):
+                for root, dirs, files in os.walk(path):
+                    for filename in files:
+                        if filename.lower().endswith(".stix2"):
```
roaree, line 55: Non-blocker: we might have some examples of using glob that could condense this into one for loop

roaree (approved):
Looks great! thanks

mvt/common/indicators.py
besendorf: happy to change that if you can show me an example.

Te-k: I think he means something like [this](https://github.com/mvt-project/mvt/blob/efe46d7b49d383d97151e03bbde3da6970f00f5f/mvt/ios/decrypt.py#L134) using the [glob library](https://docs.python.org/3/library/glob.html)

DonnchaC: > I think he means something like [this](https://github.com/mvt-project/mvt/blob/efe46d7b49d383d97151e03bbde3da6970f00f5f/mvt/ios/decrypt.py#L134) using the [glob library](https://docs.python.org/3/library/glob.html)

Yes exactly. The glob makes it easier to rationalise about what the code is doing IMO.

besendorf: I changed the code to use glob.glob instead. 
note: this is case-sensitive now. Case insensitivity with glob was only added in python3.12

PR #533, opened 2024-09-17
Update deprecated functions and other small changes

besendorf:
I updated a few deprecated functions and also made other small changes like typos in docstrings


besendorf:
Some changes are not suppoprted in python 3.9 and 3.9. What python version do we target? setup.cfg says >=3.8. Python 3.8 is eol next month. 3.9 is still supported for another year. Debian stable ships with python3.11

PR #546, opened 2024-09-30
Improve bugreport and Android dumpstate parser

DonnchaC:
This is a work-in-progress to improve our overall code for parsing bugreport files. This PR adds an initial WIP Android dumpstate file parser. These files and formats vary significantly between devices. We will need to add per-device integration tests to ensure it is working consistently.

For reference the AOSP dumpstate generation code is at https://cs.android.com/android/platform/superproject/main/+/main:frameworks/native/cmds/dumpstate/dumpstate.cpp.


ping2A:
Started to put things here:
https://github.com/CERT-EDF/dumpstate-py
https://github.com/IsMyPhonePwned/bugreport-extractor-library

PR #562, opened 2024-10-23
Improve Docker image building and add Docker info to docs

DonnchaC:


Dockerfile
```diff
@@ -130,20 +131,25 @@ COPY --from=build-libusbmuxd /build /
...
-  && apt-get install -y git python3-pip \
-  && PIP_NO_CACHE_DIR=1 pip3 install git+https://github.com/mvt-project/mvt.git@main \
-  && apt-get remove -y python3-pip git && apt-get autoremove -y \
-  && rm -rf /var/lib/apt/lists/*
+   && apt-get install -y git python3-pip \
+   && PIP_NO_CACHE_DIR=1 pip3 install --upgrade pip \
+   && PIP_NO_CACHE_DIR=1 pip3 install ./mvt \
+   && apt-get remove -y python3-pip git && apt-get autoremove -y \
+   && rm -rf /var/lib/apt/lists/* \
+   && rm -rf mvt
```
scribblemaniac, line 142: Doing this will make the repository files inaccessible in the final image, but they will still take up space in the intermediate layers. If you are willing to require BuildKit to build the Dockerfiles (the default on Linux since Febuary 2023 and easily enabled much further back), you can use the `--mount` option of `RUN` (https://docs.docker.com/reference/dockerfile/#run---mount) to pip install from the build context without saving unnecessary data in the image.

PR #566, opened 2024-10-28
WIP: Run all bugreport modules against bugreport.zip in AndroidQF extraction

DonnchaC:
We have a lot of duplication currently with separate modules for check-androidqf, check-bugreport and check-adb. 

This is starting to become difficult to maintain. This PR is a POC for running all bugreport modules automatically when running `check-androidqf` against a AndroidQF extraction containing a bugreport.zip.

Quite a few open questions here:
- [ ] How to we keep a clean interface for loading modules rather than ad-hoc code to handle loading extra modules
- [ ] This code needs to be tested against archived/zipped AQF extractions, and folders containing the extracted filles.
- [ ] How should we namespace modules to avoid name/output file collisions.
      - [ ] How can we run tests to flag name collisions or module collobering.
- [ ] We should extend this PR to also load and check backup.ab files from an AndroidQF file by loading the `check-backup` modules.
- [ ] Remove duplicate modules from `check-androidqf`.


DonnchaC:
Closing this as I have reworked this feature in #606

PR #567, opened 2024-10-28
Load Android device timezone info and add additional file modification logs

DonnchaC:
Use local timestamp for Files module timeline.

Most other Android timestamps appear to be local time. The results timeline is more useful if all the timestamps are consistent. I would prefer to use UTC, but that would mean converting all the other timestamps to UTC as well. We probably do not have sufficient information to do that accurately, especially if the device is moving between timezones.

This PR also adds new modules to load modification times for various log files in AndroidQF and bugreports which are then added to the overall timeline.


src/mvt/android/modules/androidqf/files.py
```diff
@@ -106,6 +107,12 @@ def check_indicators(self) -> None:
             # TODO: adds SHA1 and MD5 when available in MVT
 
     def run(self) -> None:
+        if timezone := self._get_device_timezone():
+            device_timezone = ZoneInfo(timezone)
+        else:
+            self.log.warning("Unable to determine device timezone, using UTC")
+            device_timezone = ZoneInfo("UTC")
```
roaree, line 114: Will look after lunch but either a `backports` or `pytz` should fix the CI failure here

Issue #410, opened 2023-11-03
Add command completion


besendorf:
Command completion for the mvt commands would definitely imporove UX. Maybe if I have some time I could implement this. But would also be a good first issue for someone trying to help the project.

r-tx:
I have completions for the mvt binaries.    Where would be the best place to put these, if they are useful?

besendorf:
this is done

PR #589, opened 2024-12-22
WIP: Add command completion

nimrod-a:
# Add command completion #410 

- Adds shell completion scripts for Bash and Zsh
- Adds documentation on activating command completion and on generating shell completions locally

For reference: [Click Docs](https://click.palletsprojects.com/en/stable/shell-completion/#enabling-completion)

## Testing
For testing, you would need to either generate your own shell completion scripts or download the ones in this commit. 
The Curl command in the docs would not work yet as the files are not available in the repo.     

## Note
As an alternative to providing the user with pre-generated shell completion scripts, one ca add the following to ~/.bashrc  / ~/.zshrc, which would result in generating the shell completion scripts every time a shell is started:
```bash
eval "$(__MVT_IOS_COMPLETE=bash_source mvt-ios)" # Example for mvt-ios in bash
```
However, this does not work properly for MVT in a virtual env, as the mvt commands are only available once the env is activated, not upon opening the shell. It would then result in an error: 
```bash
bash: mvt-ios: command not found
```
We could consider adding the `eval` instructions for users who installed MVT system-wide.


nimrod-a:
@besendorf and I discussed that it may make more sense to source the command completion files automatically. New users benefit from command completion the most, so it would be more useful if it would the user would not need extra steps for activating it. 
Any idea how we could run the following automatically in pip/docker post-install?
```bash 
 . .mvt-android-complete.bash && . .mvt-ios-complete.bash # bash 
 . .mvt-android-complete.zsh && . .mvt-ios-complete.zsh # zsh
```

roaree:
Hey @nimrod-a , thanks for the PR!

1. Installing completions into the right place automatically (vs generating a completion for each shell) seems to be something other projects avoid due to the complexity. Happy to review a PR if you find a good library for this though!

> Any idea how we could run the following automatically in pip/docker post-install?

You can definitely add completions to the Dockerfile post-install!

2. Not a big one, but I'd love to see a `mvt-ios tools generate-shell-completion` subcommand for this, instead of generating completion files by setting that magic click environment variable. Not sure that click has a public API method defined for this, so it may not be possible.

3. Let's not check the generated completions into the repo - instead we should instruct users to directly generate them into the desired location. I'm not sure the home directory is the right place for that, but I don't have any better ideas.

4. Having this in `.bashrc` :

`eval "$(__MVT_IOS_COMPLETE=bash_source mvt-ios)"`

does add significant latency the time to open a new login shell, at least on my machine. I think we should default the instructions to the second option, pre-generating the completions and sourcing them in `.bashrc` or `.zshrc`.

5. Re: problems from installing in a virtualenv, it might be worth adding an condition to check if the binaries exist on the path ala `brew`: https://docs.brew.sh/Shell-Completion . Not sure that's necessary if we default to sourcing the generated completion files instead though.

nimrod-a:
### Update: 

**1.** 
> Test enabling command completion automatically in the final docker build as written above

This would be a WIP and needs some further troubleshooting.
It may make more sense to try and solve the issue of automatic command completion "globally", regardless of the installation environment. If no general method was found we could add a fix for docker, so it's at least enabled automatically there.  
 
**2.**
> utilize [auto-click-auto](https://pypi.org/project/auto-click-auto/) to enable command completion

Upon closer look *auto-click-auto* does not seem to be the optimal solution. 
After studying their actual [implementation ](https://github.com/KAUTH/auto-click-auto/blob/main/auto_click_auto/core.py) it turns out they use `eval` internally. So basically it just adds: 
`eval "$(_MVT_IOS_COMPLETE=bash_source mvt-ios)" ` permanently to .bashrc

**3.** 
I could not really find good alternative libraries. The most known alternative library is [click-completion](https://github.com/click-contrib/click-completion), which also uses `eval` internally.

**4.**
A lot of the code from *auto-click-auto* can probably be reused and modified to work with generated scripts and not `eval`, but this would be a lot of work. It would pretty much mean making a new library for automatic Click shell completion (which could be a nice project, don't know if I have the time though) :smile:  

I will try to implement python functions for integrating command completion as a sub-command! I'll ping once there is progress on it. 

Until then: We could already merge this PR, as it just contains the updated docs for manually enabling command completion.

I could then open a new PR for automatic shell completion integration.

PR #594, opened 2025-01-06
Show characters typed in decrypt-backup; Fixes #581

### LINKED ISSUE #581: Improvement: show characters typed in decrypt-backup
It would be easier to show * instead of nothing so that people can follow the characters they typed

nimrod-a:
This PR addresses the improvement request to show characters typed in decrypt-backup as asterisks (*), so that people can follow the characters they typed. 

The [rich](https://rich.readthedocs.io/en/stable/index.html) library currently used by MVT does seem to support this feature. 
```python
password = Prompt.ask("Enter backup password", password=True) # rich library

```
The builtin password support of the library leads to the characters being completely hidden. 

Using the popular [promt_toolkit](https://python-prompt-toolkit.readthedocs.io/en/master/pages/asking_for_input.html#password-input) instead:
```python
prompt('Enter backup password: ', is_password=True) # 
# > Enter backup password: *******
``` 
When the is_password=True flag has been given, the input is replaced by asterisks (* characters).

The downside is that another library is used for password inputs. 
However, this is a fairly simple solution for the issue. 

Alternatively, one  could also try and write a custom password prompt using the rich library.

### DIFF
--- a/src/mvt/ios/cli.py
+++ b/src/mvt/ios/cli.py
@@ -8,7 +8,7 @@ import logging
 import os
 
 import click
-from rich.prompt import Prompt
+from prompt_toolkit import prompt
 
 from mvt.common.cmd_check_iocs import CmdCheckIOCS
 from mvt.common.logo import logo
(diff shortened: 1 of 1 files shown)


nimrod-a:
Understandable. Maybe the rich library will implement a customization for the password prompt, one could consider adding it as a [feature request](https://github.com/Textualize/rich/issues).

besendorf:
Discussion in the rich project regarding this feature here: https://github.com/Textualize/rich/discussions/3726

PR #597, opened 2025-01-13
Add command completion docs

nimrod-a:
As discussed in #589, it may make sense to separate the previous PR into manual and automatic command completion. 

This PR simply includes docs for manually activating command completion.
The Instructions are for Bash and Zsh.  

I will try to make a new PR for automatic command completion in the near future!


roaree:
Thanks for the contribution @nimrod-a !

PR #600, opened 2025-01-15
Add shell-completion commands

nimrod-a:
This PR is a WIP to add a `shell-completion` command to mvt-ios and mvt-android. 

As discussed in  #589, there is currently no suitable library to implement this feature without substantial performance impact.  

I have therefor created a modified fork of the most popular click autocomplete library ([auto-click-auto](https://github.com/KAUTH/auto-click-auto)), which I named [quick-click-auto](https://github.com/nimrod-a/quick-click-auto/ ).

You can read more about it [here](https://github.com/nimrod-a/quick-click-auto?tab=readme-ov-file#why-this-fork-exists)

It still needs some working on, but will hopefully do its job  in the near future. I will give an update once I have done some more testing!


nimrod-a:
#### Update: 

The `shell-completion` command should work as expected now!
I would be happy about people testing this feature :)

For testing:
```bash 
git clone -b feature/command_completion_command https://github.com/nimrod-a/mvt.git \
&& cd mvt \
&& python3 -m venv .venv && source .venv/bin/activate \
&& pip install . \
&& mvt-ios shell-completion # or mvt-android
```
Then reopen the terminal or run a new with `bash` or `zsh. 
Command completion should be enabled now. 

It works by:
- Generating the command completion scripts
- Editing the shell configuration to automatically sources the generated files to the shell. 

It should not slow down shell responsiveness as much as the `eval` implementation did, as the scripts are not generated every time the shell is opened.

besendorf:
Hey @nimrod-a if this is done and ready for review could you remove the "WIP" in the title? Could someone review this? @Te-k @roaree ?

Te-k:
I am happy to review that when it is done (and when I will have time, may take a week or two). One question we may have is do we want to have another dependency for that feature? and is this library going to be maintained if it was forked just for this project?

nimrod-a:
Hey! Sorry for the late reply. You / @roaree & @DonnchaC  probably know better then me if having another dependency is worth it for the feature. I can say that for me as a beginner MVT user it was quite useful. 

Regrading maintenance of the library: since quick-click-auto has no 3rd party dependencies besides `Click`, it will most likely only need modification in case Click changes its [Shell Completion](https://click.palletsprojects.com/en/stable/shell-completion/) implementation, which should not happen too often. 
I am quite busy with writing my thesis right now, but once I'm done I plan to make a few PRs for command completion at other projects which utilize Click, so the library will hopefully not be used only for MVT.

besendorf:
As adding a new dependency was controversial lets make it easier for users to add command completion themself so I opened a PR that adds command completion with a MVT command: https://github.com/mvt-project/mvt/pull/817

nimrod-a:
Since @besendorf created a solution without the need for a new dependency I would close this PR. Thank you @besendorf for continuing the work!

PR #615, opened 2025-03-04
Update logo.py

carlosm2:
add instructions to update mvt via pipx in case its installed via pipx


DonnchaC:
Great thank you for improving this @carlosm2

PR #617, opened 2025-03-11
Include default values when parsing tombstone protobuf files

DonnchaC:
This commit fixes bug a where default values were dropped when parsing protobuf tombstone files. This meant that important fields such as UID would be dropped with set to "0", the default value for an integer field.

Passing `include_default_values` seems to fix the issue!

### DIFF
--- a/src/mvt/android/artifacts/tombstone_crashes.py
+++ b/src/mvt/android/artifacts/tombstone_crashes.py
@@ -62,7 +62,7 @@ class TombstoneCrashResult(pydantic.BaseModel):
     process_name: Optional[str] = None
     binary_path: Optional[str] = None
     selinux_label: Optional[str] = None
-    uid: Optional[int] = None
+    uid: int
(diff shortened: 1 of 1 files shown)


src/mvt/android/artifacts/tombstone_crashes.py
```diff
@@ -62,7 +62,7 @@ class TombstoneCrashResult(pydantic.BaseModel):
     process_name: Optional[str] = None
     binary_path: Optional[str] = None
     selinux_label: Optional[str] = None
-    uid: Optional[int] = None
+    uid: int
```
enh-google, line 65: note that "arch" above shouldn't be optional either. (github won't let me comment on a context line that's outside the usual context window.)

if you're seeing that, it might be a side effect of the fact that arm32 is encoded as 0 :-( (https://android.googlesource.com/platform/system/core/+/refs/heads/master/debuggerd/proto/tombstone.proto#85)

PR #619, opened 2025-03-13
Upgrade main dockerfile runtime to ubuntu:24.04

### LINKED ISSUE #616: Sqlite3 in container to old for latest ios
With iOS 18.3.1 comes a newer version of sqlite3.
I managed to get some data out of the sms files after exporting data with the latest sqlite3 container and then read it back with the version bundled with mvt.
But I do believe that some data got lost in translation.
If anyone can point me in the direction for building a newer container I could give it a try.
Some sanitychecks about what version of sqlite3 was used to create the files compared to what version of mvt is available would be nice.

scribblemaniac:
This is to fix #616. The latest sqlite3 version in the ubuntu:22.04 image is 3.37.2, however recent versions of iOS appear to use version 3.43.2. By upgrading to ubuntu:24.04, it now uses sqlite version 3.45.1.

The build containers don't need to be updated as this is only a runtime dependency. The Dockerfile.ios file does not need to be updated because it uses a different image for its base, which already has sqlite version 3.45.3.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -103,7 +103,7 @@ RUN git clone https://github.com/libimobiledevice/usbmuxd && cd usbmuxd \
 
 
 # Create main image
-FROM ubuntu:22.04 as main
+FROM ubuntu:24.04 as main
 
 LABEL org.opencontainers.image.url="https://mvt.re"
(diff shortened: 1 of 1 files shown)


Te-k:
Sounds good!

Issue #574, opened 2024-11-28
new bug: JSON DECODE Error when loading results from logcat.txt


azeemh:
new bug shows from old phone check.

```azeem@x9:~$ mvt-android check-iocs mvtcheck


        MVT - Mobile Verification Toolkit
                https://mvt.re
                Version: 2.5.4
                Indicators updates checked recently, next automatic check in 12 hours


12:04:05 INFO     [mvt.common.cmd_check_iocs] Parsing STIX2 indicators file at path                                                                                                             
                  /home/azeem/.local/share/mvt/indicators/raw.githubusercontent.com_mvt-project_mvt-indicators_main_2023-06_01_operation_triangulation_operation_triangulation.stix2            
         INFO     [mvt.common.cmd_check_iocs] Parsing STIX2 indicators file at path                                                                                                             
                  /home/azeem/.local/share/mvt/indicators/raw.githubusercontent.com_AmnestyTech_investigations_master_2021-07-18_nso_pegasus.stix2                                              
         INFO     [mvt.common.cmd_check_iocs] Parsing STIX2 indicators file at path                                                                                                             
                  /home/azeem/.local/share/mvt/indicators/raw.githubusercontent.com_mvt-project_mvt-indicators_main_intellexa_predator_predator.stix2                                           
         INFO     [mvt.common.cmd_check_iocs] Parsing STIX2 indicators file at path                                                                                                             
                  /home/azeem/.local/share/mvt/indicators/raw.githubusercontent.com_mvt-project_mvt-indicators_main_2023-04-11_quadream_kingspawn.stix2                                         
         INFO     [mvt.common.cmd_check_iocs] Parsing STIX2 indicators file at path                                                                                                             
                  /home/azeem/.local/share/mvt/indicators/raw.githubusercontent.com_AmnestyTech_investigations_master_2023-03-29_android_campaign_malware.stix2                                 
12:04:06 INFO     [mvt.common.cmd_check_iocs] Parsing STIX2 indicators file at path                                                                                                             
                  /home/azeem/.local/share/mvt/indicators/raw.githubusercontent.com_AssoEchap_stalkerware-indicators_master_generated_stalkerware.stix2                                         
12:04:07 INFO     [mvt.common.cmd_check_iocs] Parsing STIX2 indicators file at path                                                                                                             
[...]

PR #620, opened 2025-03-27
Fix issue #574 for a module without IOCs output

ping2A:
Got the same issue as https://github.com/mvt-project/mvt/issues/574 where a module doesn't have any IOCs output.

Log the module that have no IOCs output, and skip it :)


ping2A:
By default it is breaking the current analysis, is it possible to merge if it is ok ?

Te-k:
It looks good to me, could you please just fix the syntax issues with a `ruff format src`?

Thanks!

Te-k:
Thanks!

PR #630, opened 2025-06-03
Pin NSKeyedUnarchiver version

besendorf:


Jeff-M-Brad:
Really sorry for this silly question. I was having exactly the same problem, in a Linux machine. Now that you guys have solved the issue, how can I update my system so that the error gets corrected in my machine?

Jeff-M-Brad:
> If you are installing from pypi you will need to wait for the maintainers to release a new version there. If you build from source you need to `git pull` and then `pip install -U .` again

Thank you so much! Yes, I installed from pypi, I'll have to wait, then. I don't have the knowledge yet to build from source...

Te-k:
Latest version is in pypi @Jeff-M-Brad , apologies for the delay

PR #632, opened 2025-06-08
Freeze versions and bump version

Te-k:
This PR freeze the versions of packages we are using, drops support for python < 3.10 and bumps version in order to publish 2.6.1 that fixes #630 

I will let you review it @roaree


roaree (approved):
Hey @Te-k , sorry for the wait.

I think in general our philosophy is for MVT as a distributed application - for that reason I support the idea of pinning package versions.

Given that we do distribute `mvt` as a PyPI package though, two possible snags I would consider:

1. Pinning versions might cause package resolution problems for some folks who install mvt system-wide (even though we recommend installing in a dedicated virtual environment such as with `pipx`).
2. Any other Python project that includes `mvt` as a library must not specify any other versions for these packages - while we don't officially support this usage, we do import mvt in some internal applications. This is where some more flexibility, such as pinning to major version numbers, would be more useful.

I'm happy for this to be merged and see how it goes

PR #665, opened 2025-07-19
Delete .github/workflows directory

osinaga12:
Eliminarlo


Te-k:
We don't want to remove our workflows, thanks

PR #685, opened 2025-09-01
make virustotal check also work with androidqf extractions

besendorf:


penserbjorne:
Hola! 

This is still open :)

P

PR #737, opened 2026-01-20
Tombstone: fall back to file timestamp on empty pb timestamp

besendorf:
- Allow empty protobuf timestamps to fall back to the file timestamp
- Update _parse_timestamp_string to accept a fallback
- Pass file_timestamp to its callers
- Add tests for empty and whitespace tombstone timestamps
- Adjust imports in bugreport module to include DumpsysADBState Tombstone: use file timestamp on empty pb
- skip 0 bytes tombstones and show debug warning


bitmeta69:
Thanks for working on this. I independently confirmed the zero-byte case against the released MVT v2026.7.29. A consented Android 13 bugreport contained 25 zero-byte `FS/data/tombstones/tombstone_*.pb` placeholder entries, and the Tombstones module emitted 25 `String does not contain a date` errors.

The failure also reproduces without device data:

```python
from datetime import datetime, timezone
from mvt.android.modules.bugreport.tombstones import Tombstones

Tombstones().parse_protobuf(
    "tombstone_00.pb",
    datetime(2026, 1, 1, tzinfo=timezone.utc),
    b"",
)
```

This raises `ParserError: String does not contain a date:`. The early zero-length check in this PR appears to address the exact observed case. I can validate a rebased revision against the preserved acquisition if that would be useful.

PR #772, opened 2026-04-12
Abort analysis and warn user when backup is encrypted

### LINKED ISSUE #769: Warn when trying to analyse encrypted Backup
If you capture an encrypted backup, you have to decrypt it first with `mvt-ios decrypt-backup [...]`.

However, if you forget about doing so, MVT does not warn you about it, but performs a full analysis throwing `file is not a database` many times.

It would be so cool if the `mvt.ios.modules.backup.manifest` module could just stop the whole analysis (and suggest decryption) when it detects unexpected format of the `Manifest.db`.

besendorf:
Fixes #769

## Summary

- Adds a new `EncryptedBackupError` exception to `mvt.common.module`
- In `Manifest.run()`, catches `sqlite3.DatabaseError` when querying `Manifest.db` — the error that occurs when the database is encrypted — and raises `EncryptedBackupError`
- In `Command.run()`, catches `EncryptedBackupError` from `run_module()`, logs a clear critical message telling the user to decrypt the backup first using `mvt-ios decrypt-backup`, and stops the entire analysis immediately

Previously, running `check-backup` against an encrypted backup would silently continue through all modules, producing a flood of confusing "file is not a database" errors. Now the analysis aborts early with a single actionable message.

### DIFF
--- a/src/mvt/common/command.py
+++ b/src/mvt/common/command.py
@@ -11,7 +11,7 @@ from datetime import datetime
 from typing import Optional
 
 from mvt.common.indicators import Indicators
-from mvt.common.module import MVTModule, run_module, save_timeline
+from mvt.common.module import EncryptedBackupError, MVTModule, run_module, save_timeline
 from mvt.common.utils import (
     convert_datetime_to_iso,
     generate_hashes_from_path,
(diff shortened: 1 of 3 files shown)


Max-RSF:
Awesome! It works great for me and protects from so much confusion when starting with MVT :tada:

PR #798, opened 2026-05-15
Fix timeline cross run contamination

### LINKED ISSUE #797: Timeline append mode causes cross run contamination
`save_timeline()` uses append mode (a+), so reusing an output directory silently merges old and new results. It should use write mode (w) since the timeline is already fully built in memory before saving.

tesitura:
Closes #797 
Switch `save_timeline()` for `timeline.csv` and `alerts_timeline.csv` from append to write mode. Both timelines are fully built in memory before writing, so append only caused silent data contamination across runs.

### DIFF
--- a/src/mvt/common/alerts.py
+++ b/src/mvt/common/alerts.py
@@ -219,7 +219,7 @@ class AlertStore:
         return alerts
 
     def save_timeline(self, timeline_path: str) -> None:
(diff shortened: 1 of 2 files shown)


besendorf:
Lgtm

rjustice439-pixel (approved):
Oi

PR #811, opened 2026-06-09
intrusion_logs: alert on certificate events and run heuristics without IOCs

kavanista:
## Summary

- run `SecurityEvent` heuristic alerts independently of IOC availability, so
  they fire on a default run with no indicator set loaded
- surface `cert_authority_installed` and `cert_validation_failure` through the
  alert store at medium severity, matching their sibling events
- add regression coverage for the certificate alerts and for heuristics firing
  with no indicators loaded

## Root cause

`SecurityEvent.check_indicators()` returned immediately when no indicator set
was loaded. Because `check_indicators()` is always called, a default
`check-intrusion-logs` run produced none of the module's heuristic alerts:
key-integrity violations, wipe failures, crypto self-test failures, and the
certificate events. Separately, `cert_authority_installed` and
`cert_validation_failure` only emitted `log.warning` while sibling events
called `alertstore.medium`, so a root CA installation (potential MITM) and a
certificate validation failure never reached the alert report.

This mirrors the accessibility fix in #807: heuristic alerts now run
independently of IOC availability, while IOC matching stays gated on a loaded
indicator set.

### DIFF
--- a/src/mvt/android/modules/intrusion_logs/security_event.py
+++ b/src/mvt/android/modules/intrusion_logs/security_event.py
@@ -422,16 +422,22 @@ class SecurityEvent(IntrusionLogsModule):
                 result,
             )
 
-        # Flag certificate authority installations (potential MITM)
+        # Flag successful certificate authority installations (potential
+        # MITM); a missing success field is treated as installed
         if "cert_authority_installed" in result:
             cert_info = result["cert_authority_installed"]
-            self.alertstore.medium(
-                "Certificate authority installed: "
-                f"{cert_info.get('subject', 'unknown')} "
-                f"(success: {cert_info.get('success', 'unknown')})",
-                result.get("timestamp") or "",
-                result,
-            )
+            if cert_info.get("success", True):
+                self.alertstore.medium(
+                    "Certificate authority installed: "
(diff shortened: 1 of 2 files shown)


besendorf:
Hi,
thanks for this PR. I think I found one issue with it:

Failed CA installations produce false alerts
    [security_event.py#L425-L434](https://github.com/mvt-project/mvt/blob/1c6ddc67a67a7cd35fb47a75be0704c7c576eefc/src/mvt/android/modules/intrusion_logs/security_event.py#L425-L434)
emits “Certificate authority installed” regardless of success. Android defines 0 as failed and 1 as succeeded
[documentation](https://developer.android.com/reference/android/app/admin/SecurityLog#TAG_CERT_AUTHORITY_INSTALLED). Please Gate the alert on success and add a failed-install regression test.

kavanista:
Good catch, thanks! Fixed in 4fb8614, the certificate authority install alert is now gated on the event's `success` field, so failed install attempts no longer produce false alerts. Also added a regression test covering the failed install case in test_intrusion_logs.py

kavanista:
@besendorf need anything else here ?

PR #812, opened 2026-06-10
fix: use exact package match in bugreport receiver check

### LINKED ISSUE #803: False positive: `check_receiver_prefix` does unanchored substring match, triggers CRITICAL on AOSP `com.android.phone` for the TheOneSpy / `com.android.services` indicator
## Bug

In `check-bugreport` / `check-androidqf` runs, `dumpsys_receivers_detected.json` reports a CRITICAL match:

```json
{
  "message": "Found a known suspicious receiver with name \"com.android.phone/com.android.services.telephony.sip.SipIncomingCallReceiver\" matching indicators from \"TheOneSpy\"",
  "event": {"package_name": "com.android.phone",
            "receiver":     "com.android.phone/com.android.services.telephony.sip.SipIncomingCallReceiver"},
  "matched_indicator": {"value": "com.android.services", "type": "app_ids", "name": "TheOneSpy"}
}
```

The matched `com.android.phone` is the AOSP SIP telephony receiver, not TheOneSpy. The indicator value `com.android.services` is a real TheOneSpy masquerade package id (AssoEchap data is correct) — but it appears here only as a substring of `com.android.services.telephony.sip.SipIncomingCallReceiver` inside system package `com.android.phone`.

Reproduces on MVT `2026.5.12` against any device still shipping SIP telephony in `com.android.phone` (Android ≤ 13 stock, plus OEM ROMs that retained SIP after the API 34 removal — SIP was deprecated in API 31 and removed in API 34). Verified on HUAWEI STK-L21 / EMUI 10.

## Where

Introduced by #721 (merged 2026-01-10). Two files:

`src/mvt/android/modules/bugreport/dumpsys_receivers.py:38-52` — the only caller of `check_receiver_prefix` in the codebase, and the only Android module that compares an indicator against the full `package/class` string instead of `package_name`:

```python
receiver_name = self.results[result][0]["receiver"]
ioc_match = self.indicators.check_receiver_prefix(receiver_name)
```

`src/mvt/common/indicators.py:730-751` — unanchored substring containment, despite the name `check_receiver_prefix`:

```python
for ioc in self.get_iocs("app_ids"):
    if ioc.value.lower() in receiver_name.lower():    # ← unanchored
        return IndicatorMatch(ioc=ioc, ...)
```

Every other Android artifact (≈14 callsites) uses `check_app_id` (`indicators.py:709-728`), which does exact equality against `package_name`. That code path would not produce this FP.

Notably, the sibling artifact module `src/mvt/android/artifacts/dumpsys_receivers.py:53` already does the right thing — it calls `check_app_id(receiver["package_name"])`. Only the bugreport wrapper in `modules/bugreport/dumpsys_receivers.py` was switched to the new `check_receiver_prefix` path in #721, so the two receiver-matching paths in the codebase now disagree.

## Fix

Smallest change — drop `check_receiver_prefix` and call `check_app_id` on the package portion:

```python
# modules/bugreport/dumpsys_receivers.py
package_name = receiver_name.split("/", 1)[0]
ioc_match = self.indicators.check_app_id(package_name)
```

If checking class paths is intentional (the goal of #721), anchor on a namespace boundary instead of using `in`:

```python
def check_receiver_prefix(self, receiver_name: str):
    if not receiver_name:
        return None
    lower = receiver_name.lower()
    pkg = lower.split("/", 1)[0]
    for ioc in self.get_iocs("app_ids"):
        v = ioc.value.lower()
        if pkg == v or lower.startswith(v + "."):
            return IndicatorMatch(ioc=ioc, ...)
    return None
```

Under either fix, the canonical FP disappears (`com.android.phone` ≠ `com.android.services`, and `com.android.services` is not at a namespace boundary inside the SIP class path), while a real `com.android.services/...` receiver still matches.

Regression test:

```python
def test_check_receiver_prefix_does_not_match_aosp_namespace_substring():
    ind = ... # TheOneSpy / app_ids: ["com.android.services"]
    assert ind.check_receiver_prefix(
        "com.android.phone/com.android.services.telephony.sip.SipIncomingCallReceiver"
    ) is None
    assert ind.check_receiver_prefix(
        "com.android.services/com.example.SomeReceiver"
    ) is not None
```

Happy to send a PR for option 1 or 2 — preference welcome.

DevenDucommun:
The check_receiver_prefix method used unanchored substring matching, causing false CRITICAL alerts when a legitimate AOSP receiver class path contained an indicator value as a substring (e.g. TheOneSpy's "com.android.services" matching inside com.android.phone's SIP telephony receiver).

Replace with check_app_id on the extracted package name, consistent with the artifact module's approach. Remove the now-unused check_receiver_prefix method.

Fixes #803

AI assisted

### DIFF
--- a/src/mvt/android/modules/bugreport/dumpsys_receivers.py
+++ b/src/mvt/android/modules/bugreport/dumpsys_receivers.py
@@ -39,9 +39,9 @@ class DumpsysReceivers(DumpsysReceiversArtifact, BugReportModule):
         for result in self.results:
             if self.indicators:
                 receiver_name = self.results[result][0]["receiver"]
+                package_name = receiver_name.split("/", 1)[0]
 
-                # return IoC if the stix2 process name a substring of the receiver name
-                ioc_match = self.indicators.check_receiver_prefix(receiver_name)
+                ioc_match = self.indicators.check_app_id(package_name)
                 if ioc_match:
                     self.alertstore.critical(
                         ioc_match.message,
--- a/src/mvt/common/indicators.py
+++ b/src/mvt/common/indicators.py
@@ -727,29 +727,6 @@ class Indicators:
 
         return None
 
-    def check_receiver_prefix(
-        self, receiver_name: str
-    ) -> Optional[IndicatorMatch]:
-        """Check the provided receiver name against the list of indicators.
-        An IoC match is detected when a substring of the receiver matches the indicator.
-
-        :param receiver_name: Receiver name to check against app ID indicators
-        :type receiver_name: str
-        :returns: IndicatorMatch if matched, otherwise None
-
-        """
-        if not receiver_name:
-            return None
-
-        for ioc in self.get_iocs("app_ids"):
-            if ioc.value.lower() in receiver_name.lower():
-                return IndicatorMatch(
-                    ioc=ioc,
-                    message=f'Found a known suspicious receiver with name "{receiver_name}" matching indicators from "{ioc.name}"',
-                )
-
-        return None
-
     def check_android_property_name(
         self, property_name: str
     ) -> Optional[IndicatorMatch]:
--- a/tests/common/test_indicators.py
+++ b/tests/common/test_indicators.py
@@ -172,6 +172,15 @@ class TestIndicators:
         assert ind.check_android_property_name("sys.foobar")
         assert ind.check_android_property_name("sys.soundsokay") is None
 
+    def test_check_app_id_no_substring_match(self, indicator_file):
+        ind = Indicators(log=logging)
+        ind.load_indicators_files([indicator_file], load_default=False)
+        ind.ioc_collections[0]["app_ids"].append("com.android.services")
+
+        assert ind.check_app_id("com.android.services") is not None
+        assert ind.check_app_id("com.android.phone") is None
+        assert ind.check_app_id("com.android.services.telephony") is None
+
     def test_env_stix(self, indicator_file):
         os.environ["MVT_STIX2"] = indicator_file
         settings.__init__()  # Reset settings


viktor3002:
Hi @DevenDucommun, thanks for the PR! This sounds all very good and plausible, but I'm having trouble reproducing and testing the issue on my side. 
Do you maybe have a bugreport you can share with us that triggers the bug? If yes, feel free to send a link to mail@lab.rsf.org. I would then test it and merge the PR.

DevenDucommun:
Hi @viktor3002, thanks for taking a look.

I don't have a real-world bugreport I can share, but the false positive doesn't need one. It only requires a receiver whose class path contains `com.android.services` as a substring while the package is something else. The AOSP SIP telephony receiver in `com.android.phone` does exactly that, and it ships on plenty of Android <= 13 builds (SIP was deprecated in API 31 and removed in API 34, so older stock and OEM ROMs that kept it still have it).

So I put together a minimal synthetic bugreport that triggers it. Two files in a zip:

`main_entry.txt`:
```
dumpstate.txt
```

`dumpstate.txt` (the relevant part):
```
DUMP OF SERVICE package:

Receiver Resolver Table:
  Non-Data Actions:
      android.intent.action.PHONE_STATE:
        37c9e2a com.android.phone/com.android.services.telephony.sip.SipIncomingCallReceiver filter 9f02bb1
```

And a one-indicator STIX2 file with TheOneSpy's real `com.android.services` app id:
```json
{ "pattern": "[app:id='com.android.services']", "pattern_type": "stix" }
```

Running `mvt-android check-bugreport -i theonespy.stix2 -o out bugreport.zip`:

- On current `main`, this produces `dumpsys_receivers_detected.json` with a CRITICAL alert, message identical to the one in #803: `Found a known suspicious receiver with name "com.android.phone/com.android.services.telephony.sip.SipIncomingCallReceiver" matching indicators from "TheOneSpy"`.
- With this PR applied, the receiver is still parsed, but no detection fires.

I'll also email the ready-made zip and stix2 file to mail@lab.rsf.org so you can run it directly. Happy to fold this into a regression test fixture under `tests/artifacts/android_data/bugreport/` if you'd like.

(For transparency: the patch and this analysis were done with AI assistance, as noted in the PR body.)

DevenDucommun:
This closed automatically when I deleted my old fork during some personal repo cleanup. I didn't realize deleting a fork closes its open PRs. The change is unchanged, so I've re-filed it as #840. Sorry for the noise.

PR #841, opened 2026-07-16
Add support for parsing Protobuf-formatted artifacts

TheZ3ro:
Why?
Because the current JSON format used by MVT/AndroidQF do not directly allow for stream-read/write.

This becomes a problem when such JSON files must be read/written on devices with limited RAM.
For example the [Files](https://github.com/mvt-project/androidqf/blob/f618b7ae6775f37a2cda88f09a7b634fc7c6fef2/modules/files.go#L14) module of AndroidQF, reads all the data in memory and then store it in the acquisition archive.
On real-world devices with a lot of files this can become a problem.

Bugbane currenly produce size-delimited Protobuf artifacts instead of JSON to reduce RAM footprint, since they allow for partial and sequential read/write operation out-of-the-box. (write a single entry without having the whole array loaded in memory)

This PR adds upstream support for Protobuf-formatted artifacts.
JSON files always get the priority. If a JSON file is found, it is read and parsed. If no JSON files are found, the Protobuf counterpart are searched and loaded instead.
Protobuf parsing uses the betterproto2 library that is already used for parsing Tombstone artifacts.

This PR also add tests for all the relevant Protobuf artifact parsing.

Happy to make fixes or modifications if needed :)


src/mvt/android/parsers/proto/__init__.py
```diff
@@ -0,0 +1,603 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
besendorf, line 1: **[P1] Keep the repository mypy check passing**

This generated package is now discovered under both `src.mvt.android.parsers.proto` and `mvt.android.parsers.proto`. Running `make check` at this PR head passes Ruff but fails immediately in mypy with `Source file found twice under different module names`; current `main` passes the same mypy command. Please adjust the generated package layout/typing marker or mypy packaging so `make mypy` succeeds.

src/mvt/android/modules/androidqf/mounts.py
```diff
@@ -42,30 +51,37 @@ def run(self) -> None:
...
+                self.log.error("Failed to parse JSON mount information: %s", exc)
+                return
 
-        try:
-            data = self._get_file_content(mount_files[0]).decode(
-                "utf-8", errors="replace"
-            )
-        except Exception as exc:
-            self.log.error("Failed to read mount information file: %s", exc)
+        mount_files = self._get_files_by_pattern("*/mounts.pb")
```
besendorf, line 66: **[P2] Preserve JSON priority when both formats exist**

This protobuf lookup runs unconditionally after a successfully parsed `mounts.json`, so `mount_data` is overwritten whenever both files exist. `root_binaries.py` has the same unconditional overwrite. That contradicts the PR contract that JSON takes priority and can make MVT analyze the lower-priority artifact; the new fixture contains both formats with identical values, which masks the regression. Please only try protobuf when no JSON file was found.

src/mvt/android/parsers/protobuf_parsers.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def parse_string_records(data: bytes) -> list[str]:
+    """Parse root_binaries.pb or mounts.pb into a list of string values."""
+    return [record.value for record in _parse_delimited_messages(data, StringRecord)]
+
+
+def files_record_to_dict(record: FilesRecord) -> dict[str, Any]:
+    """Map protobuf file metadata to the AndroidQF files.json result shape."""
+    modified_time = int(record.mtime) if record.mtime else 0
```
besendorf: **[P2] Preserve fractional file timestamps**

Bugbane collects `find %T@` into a `Double`, but converting it to `int` here discards sub-second precision before MVT builds its forensic timeline. `convert_datetime_to_iso()` already handles fractional timestamps, so please retain the floating-point value instead of truncating it. Producer: https://github.com/osservatorionessuno/bugbane/blob/b6560649971b25fc60e9d9dbef40bf8172735d72/app/src/main/java/org/osservatorionessuno/qf/modules/Files.kt

besendorf (changes requested):
I found three blocking correctness and validation issues. The full test suite passes (175 passed, 1 skipped), but the repository mypy check fails on this branch.

besendorf:
Hi,
I get your use case and I would like to support it. However Im not sure if protobuf is actually needed for the reasons you mentioned. We could also use JSON lines aka new line delimited json or even stream regular json. For example if you have large arrays in json you can just stream each element one by one. Kotlin even has support for streaming regular json in [JsonWriter](https://developer.android.com/reference/kotlin/android/util/JsonWriter)

  The streaming methods are:

  - beginArray() (https://developer.android.com/reference/kotlin/android/util/JsonWriter#beginArray())
  - beginObject() (https://developer.android.com/reference/kotlin/android/util/JsonWriter#beginObject())
  - name(String) (https://developer.android.com/reference/kotlin/android/util/JsonWriter#name(java.lang.String))
  - value(...) (https://developer.android.com/reference/kotlin/android/util/JsonWriter#value(boolean))
  - endObject() (https://developer.android.com/reference/kotlin/android/util/JsonWriter#endObject())
  - endArray() (https://developer.android.com/reference/kotlin/android/util/JsonWriter#endArray())
  - flush() (https://developer.android.com/reference/kotlin/android/util/JsonWriter#flush())

I also understand that you already produce protobuf files though. I would wait until a few other people have voiced an opinion.

src/mvt/android/parsers/protobuf_parsers.py
TheZ3ro: great catch, I though `convert_datetime_to_iso` expected UNIX int timestamps. fixed

src/mvt/android/modules/androidqf/mounts.py
TheZ3ro: right, I've added a check to confirm that JSON parsing didnt load anything before proceeding with protobuf.
added also test to confirm it.

src/mvt/android/parsers/proto/__init__.py
TheZ3ro: I fixed the error, however I think mypy is currently configured improperly.

currently pyproject.toml sets `packages = "src"`.
however, maybe a more appropriate configuration would be setting `mypy_path = "src"` and then `packages = "mvt"` so that the root package name will be correctly set to `mvt`.

TheZ3ro:
Hi @besendorf,
many thanks, I've addressed your comments :)

I agree that it is indeed possible to use JSON or JSONL.
The reasons that led us to adopt Protobuf for our niche use case are multiple.

Protobuf is slightly faster (less characters to write/read compared to JSON key-values), has a lower size (10x, however after compression the difference is negligible. but MVT only supports uncompressed AndroidQF parsing)

Finally, Protobuf allows for a fixed standard/source of truth. You can use that in Kotlin/Python/Go and everyone will benefit (AndroidQF/MVT/Bugbane would all adhere to the same format)

TheZ3ro:
Any updates on this? :)

besendorf:
@DonnchaC needs to decide whether we want to support this

DonnchaC:
Hi @TheZ3ro, thanks so much for all your work on Bugbane and for opening this PR.

I think I have a good sense of the constraints you have 1) minimizing RAM usage during generation and parsing 2) having a structured and defined output format to ensure compatibility between AndroidQF-like tools (Bugbane etc) and the MVT parsers.

On this specific Profobuf proposal: I would lean towards updating updating AndroidQF and MVT to generate and parse (compressed) JSONL. We use JSON for other artifacts, and I believe JSONL should allow streaming reads and writes similar to Protobuf. I know Protobuf also enforces a schema which is valuable, but I'd propose we enforce schema via a community-agreed output format spec, rather than having a unique case for Bugbane.    

I think both the schema validation and RAM/JSONL changes can both be solved best with a new AndroidQF archive specification. This will also allow other AndroidQF-like implementations (WebUSB AndroidQF, a potential desktop GUI client etc) to create a consistent and interoperable artifacts.

We've got a draft AndroidQF output specification which I'd be happy to share and get your thoughts on. I want to be sure it meets the needs of Bugbane and other community projects.

Once the AndroidQF schema is finalised we would 

1) update AndroidQF to prepare achives using the new output structure with JSONL etc. 
2) Update MVT to handle the new AndroidQF format 
3) publish the version schema and 
4) publish a AndroidQF test suite and archive validator to allow other implementations to verify compatibility.

If that seems like a workable approach for you, then I'd suggest we close this PR for now and pick this up again once we've got the new AndroidQF specification ready

DonnchaC:
I'll close for now pending further discussion of the best output format and specification.

TheZ3ro:
Hi @DonnchaC,
thank you for going through this and providing a nice feedback, it's really appreciated.

Correct, the main constrains and goals on our side are computation usage (RAM, CPU and output size) and compatibility with the whole MVT environment and already established tools.
Sadly the current JSON format is too demanding for low-/medium-end phones from our benchmark, that's why we decided to go with Protobuf and contribute to MVT in order to add support for it.
Forking or creating new tools is a non-goal.

I agree that a formal specification is needed and this was partially what we were trying to pursue (formalizing the current acquisition format) so we are more than happy to hear that you are working on it and you are shaping it to include most of the relevant use cases.
I would love to see and contribute on the specification once you have a pre-final shape.

I must admit that I still prefer Protobuf to JSONL but it's subjective and I definitely don't want to turn this into a format war :D
I'm happy to close this PR and follow up to adopt the new specification in Bugbane!

PR #846, opened 2026-07-26
Scan Safari profile databases for history and browser state

sapran:
## Problem

Safari profiles (iOS 17 and later) keep their own databases under
`Library/Safari/Profiles/<UUID>/`. `SafariHistory` and `SafariBrowserState`
only ever looked at the default profile's `Library/Safari/History.db` and
`Library/Safari/BrowserState.db`.

The result is a silent gap rather than an error: on a device where browsing
happens inside a profile, MVT skips that history entirely and still reports
`The SafariHistory module produced no detections!`. An indicator that was only
ever visited within a profile goes unnoticed, and nothing in the output hints
that a database was missed.

I ran into this while checking an encrypted iOS 26.5.2 backup that contained 14
per-profile `History.db` files under
`AppDomain-com.apple.mobilesafari :: Library/Safari/Profiles/<UUID>/History.db`,
none of which MVT opened.

## Fix

Both modules now also match `Library/Safari/Profiles/*/` for backups and for
filesystem dumps.

No helper changes were needed:

- `_get_backup_files_from_manifest()` already translates `*` into a SQL `LIKE`
  wildcard (`base.py`), and `%` spans `/` in `LIKE`, so the pattern matches at
  any profile UUID.
- `_get_fs_files_from_patterns()` already globs.

The lookup is done per pattern rather than with one loose pattern on purpose.
A single `Library/Safari/*History.db` would work, but SQLite's `LIKE` is
case-insensitive for ASCII, so it would also pull in unrelated files such as
`Library/com.apple.internal.ck/history.db`.

## A note on scope

The per-profile `History.db` layout is confirmed against a real device. I did
**not** observe a per-profile `BrowserState.db` in that backup — its profiles
were created but never browsed in, so absence there is not strong evidence
either way. I applied the same pattern to `SafariBrowserState` for symmetry: a
glob that matches nothing costs one extra `Manifest.db` query, whereas leaving
it out preserves a known-shaped blind spot. Happy to drop that half if you'd
rather only ship the verified path.

## Tests

Adds `tests/ios_backup/test_safari_history.py` (the module had no test
coverage before) and one case to `test_safari_browserstate.py`. Fixtures build
a backup and a filesystem dump that each contain a default profile plus a named
profile, so the tests exercise both lookup paths.

All 5 new tests fail on `main` and pass with this change:

```
FAILED test_safari_history.py::test_parsing_backup_with_profile   - assert 1 == 2
FAILED test_safari_history.py::test_parsing_fs_dump_with_profile  - assert 1 == 2
FAILED test_safari_history.py::test_detection_in_profile          - assert 0 == 1
FAILED test_safari_browserstate.py::test_parsing_backup_with_profile - assert 1 == 2
```

`test_detection_in_profile` is the one that matters: it puts an IOC-matching URL
in profile history and asserts MVT alerts on it. Without the fix that run
reports no detections.

`make ruff`, `make mypy` and the full `pytest` suite pass.

One shared helper (`add_backup_manifest_entry`) was added to `tests/utils.py`,
since both test modules need to register a file in a copied backup's
`Manifest.db`. It flips the copy out of WAL mode after inserting, because MVT
opens backup databases with `immutable=1`, which ignores any `-wal` sidecar.

### DIFF
--- a/src/mvt/ios/modules/mixed/safari_browserstate.py
+++ b/src/mvt/ios/modules/mixed/safari_browserstate.py
@@ -19,10 +19,17 @@ from mvt.common.utils import convert_mactime_to_iso, keys_bytes_to_string
 
 from ..base import IOSExtraction
 
-SAFARI_BROWSER_STATE_BACKUP_RELPATH = "Library/Safari/BrowserState.db"
+# Safari profiles (iOS 17 and later) keep their per-profile databases under
+# Library/Safari/Profiles/<UUID>/, separate from the default profile's.
+SAFARI_BROWSER_STATE_BACKUP_RELPATHS = [
+    "Library/Safari/BrowserState.db",
+    "Library/Safari/Profiles/*/BrowserState.db",
+]
 SAFARI_BROWSER_STATE_ROOT_PATHS = [
     "private/var/mobile/Library/Safari/BrowserState.db",
+    "private/var/mobile/Library/Safari/Profiles/*/BrowserState.db",
     "private/var/mobile/Containers/Data/Application/*/Library/Safari/BrowserState.db",
+    "private/var/mobile/Containers/Data/Application/*/Library/Safari/Profiles/*/BrowserState.db",
 ]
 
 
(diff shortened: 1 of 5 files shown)


tomritzow (commented):
[]()

PR #890, opened 2026-08-25
Fix getprop parsing of multi-line property values

va-resident:
`GetProp.parse()` matches `[name]: [value]` line by line, so a value containing newlines never matches and the property is dropped — with no log line, only the usual "Extracted N Android system properties". `persist.sys.boot.reason.history` is multi-line by construction (one boot per line), so it is missing on nearly every device. It is the device's own record of why it restarted, with an epoch timestamp per event (`reboot,userrequested`, `reboot,ota`, `recovery`, `shutdown,battery`, `kernel_panic`, `watchdog`).

Raw:

```
[persist.sys.boot.reason.history]: [reboot,userrequested,mainline_update,1782907395
reboot,userrequested,1782582256
shutdown,userrequested,1781742871]
```

Fix: match the whole section once, with the value closed by the first `]` that ends a line.

```python
rxp = re.compile(r"^[ \t]*\[([^\]\n]+)\]: \[(.*?)\][ \t\r]*$", re.MULTILINE | re.DOTALL)
```

The value stays one newline-joined string, so the record shape `{"name", "value"}` is unchanged for consumers. Happy to switch to a list if you prefer.

**Relation to #871.** #871 already rewrites this regex as `^\[([^]]+)\]: \[(.*)\]$`, which fixes the empty-value (`[name]: []`) and bracket-in-value cases on its own — those are its fixes, not this PR's, and this PR is deliberately narrowed to the multi-line case, which a line-by-line match cannot capture. Marked as a draft so it does not compete for review or add a conflict to #871: after #871 lands I will rebase this onto it, keeping its `[^]]+` name group. The same suggestion is posted inline on #871 if you would rather fold it in there and close this.

Checked in the worktree: `ruff check`, `mypy`, `pytest tests` all pass; new `test_parsing_multiline_value` with a three-line property in the shared `getprop.txt` fixture (count assertion 13 → 14).

### DIFF
--- a/src/mvt/android/artifacts/getprop.py
+++ b/src/mvt/android/artifacts/getprop.py
@@ -28,19 +28,16 @@ INTERESTING_PROPERTIES = [
 class GetProp(AndroidArtifact):
     def parse(self, entry: str) -> None:
         self.results: List[Dict[str, str]] = []
-        rxp = re.compile(r"\[(.+?)\]: \[(.+?)\]")
+        # A property value may span several lines: persist.sys.boot.reason.history
+        # prints one boot per line. Matching the whole section instead of line by
+        # line lets the value run to the first closing bracket that ends a line.
+        rxp = re.compile(
+            r"^[ \t]*\[([^\]\n]+)\]: \[(.*?)\][ \t\r]*$",
+            re.MULTILINE | re.DOTALL,
+        )
 
-        for line in entry.splitlines():
-            line = line.strip()
-            if line == "":
-                continue
-
-            matches = re.findall(rxp, line)
-            if not matches or len(matches[0]) != 2:
-                continue
-
-            prop_entry = {"name": matches[0][0], "value": matches[0][1]}
-            self.results.append(prop_entry)
+        for name, value in rxp.findall(entry):
+            self.results.append({"name": name, "value": value})
 
     def get_device_timezone(self) -> str | None:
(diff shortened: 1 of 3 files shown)


vd-beep-beep:
Overlap note, so this does not sit in the way of #871: that PR also rewrites this same regex, as `^\[([^]]+)\]: \[(.*)\]$` matched line by line. That version already covers two of the three shapes described above — an empty value `[name]: []`, and a value containing `]` (greedy `.*` up to the bracket at end of line) — so the only part unique to this PR is the multi-line value, which a line-by-line match cannot capture by construction: `persist.sys.boot.reason.history` prints one boot per line, so the property is dropped entirely on nearly every device.

Whichever is easier for you:

* I narrow this PR to the multi-line case only, rebased on top of #871 (a `MULTILINE | DOTALL` regex over the whole section, keeping #871's name group), or
* I close this and you fold the multi-line variant into #871 directly.

Happy either way — just say which, and I will also add the `persist.sys.boot.reason.history` fixture case wherever it ends up.

va-resident:
Closing this in favour of #871, which now carries the same change — applied from the inline suggestion at https://github.com/mvt-project/mvt/pull/871#discussion_r3855437320, with tests for the multi-line and empty-value cases added there. Nothing left here that #871 does not cover. Thanks for picking it up.

PR #909, opened 2026-09-01
Automate weekly calendar version bumps

besendorf:


DonnchaC:
I've merged #910 which adds automated PyPI deploys too,

PR #912, opened 2026-09-03
Do not keep the settings row's own metadata inside the value

va-resident:
## The bug

`dumpsys settings` prints each row's own metadata after the value:

```
value:1 default:1 defaultSystemSet:true isValuePreservedInRestore:true
value:4 notPreservedInRestore
```

The inline strip in `Settings.parse()` matched one shape only —
`default:X defaultSystemSet:true|false` at end of line — so any other
combination stayed inside the parsed value. `check_indicators()` compares that
string to `safe_value` literally, so a dangerous setting sitting at its safe
value is reported as suspicious.

## The fix

Capture the value to end of line, then split the metadata off at the **first**
metadata key, not the last — a metadata value can itself contain spaces
(`value:Galaxy Z Flip6 default:Galaxy Z Flip6 defaultSystemSet:true`).

## Impact, measured on 30 bug reports (12 vendors, Android 10-16)

| | |
|---|---|
| settings rows | 80527, of which 35874 carry trailing metadata |
| rows keeping it inside the value | 19056, on 19 of 30 devices |
| dangerous-settings alerts before / after | **60 / 8** |
| removed / added | 52 / 0 (parsed key sets identical on all 30) |

The 8 survivors were confirmed against the raw rows. Most frequent false
positive: `accessibility_enabled`, reported as "enabled accessibility services"
on 15 of 30 devices whose value is `0`.

Only `check-bugreport` is affected — `AQFSettings` parses androidqf's
`settings_*.txt` as `key=value` and never reaches this parser.

Split-point safety: across those 35874 rows, no metadata key occurs twice after
`value:` and no remainder ends in anything other than a metadata token, so the
"value containing ` default:`" case does not occur. The value is now also
`.strip()`ed, which affects 10 rows on 7 devices (`value: [ UN UN UN ]`).

## Test plan

- [x] 6 new tests in `tests/android/test_artifact_settings_metadata.py`
- [x] `ruff check`, `ruff format --check`, `mypy`: clean
- [x] `pytest`: full suite green, and re-measured on the 30 bug reports above

### DIFF
--- a/src/mvt/android/artifacts/settings.py
+++ b/src/mvt/android/artifacts/settings.py
@@ -7,6 +7,15 @@ import re
 
 from .artifact import AndroidArtifact
 
+# The row's own metadata follows the value ("value:1 default:1
+# defaultSystemSet:true isValuePreservedInRestore:true"). It starts at the
+# FIRST of these keys, not the last: a metadata value can itself contain
+# spaces ("value:Galaxy Z Flip6 default:Galaxy Z Flip6 ...").
+TRAILING_META_RXP = re.compile(
+    r"\s+(?:default|defaultSystemSet|tag|isValuePreservedInRestore):"
+    r"|\s+(?:not)?[Pp]reservedInRestore\b"
+)
+
 ANDROID_DANGEROUS_SETTINGS = [
     {
         "description": "disabled Google Play Services apps verification",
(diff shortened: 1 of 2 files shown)


DonnchaC:
I've resolved this issue in https://github.com/mvt-project/mvt/pull/914

PR #914, opened 2026-09-04
Parse dumpsys settings as per-record results

DonnchaC:
`Settings.parse` reads the `dumpsys settings` section with one regex per line. That is wrong in two ways which cost records and corrupt values, and it throws away most of what the dump actually contains.

## Defect 1 — values are truncated, and metadata leaks into them

`dumpsys` prints one record per settings row, but a value is free-form: it can contain spaces and run over many lines, and the `default:` / `defaultSystemSet:` metadata is printed after it, so it can end up on a later line than the value starts on. A per-line regex sees none of that.

The bugreport test fixture now carries a `DUMP OF SERVICE settings:` section modelled on the real output. Before/after on that fixture:

| setting | before | after |
| --- | --- | --- |
| `namespace_one/blocked_components` | `com.android.settings,com.android.vending,` | all four packages, over three lines |
| `namespace_one/allowed_packages` | `com.example.messaging,` | `com.example.messaging,\ncom.example.chat` |
| `namespace_one/streaming_blocked_components` | `com.example.dialer,com.example.camera default:com.android.settings,` | `com.example.dialer,com.example.camera` |
| `widget_instance_data` | `{` | the whole JSON document |

The third row is the metadata leak: `defaultSystemSet:` falls on a later line, so the optional tail of the old regex cannot match and ` default:...` stays in the value.

## Defect 2 — records are silently dropped

Results were a `dict` of `dict`s keyed by `namespace -> setting name`, which collapses a name recorded more than once, and the regex required a `pkg:` field which `dumpsys` omits when no package is recorded. On the same fixture the old parser keeps 9 of the 11 rows: the second `widget_instance_data` row (same name, different `_id`) overwrites the first, and `hidden_api_blacklist_exemptions` (no `pkg:`) never matches at all.

## What is parsed now

A line loop accumulates one record at a time, up to the next row, heading or blank line, and splits its `key:value` fields once the whole record has been read. The value and the default are free-form, so each field runs to the next key dumpsys printed. A blank line closes a record because dumpsys prints one after every namespace block and after a change history, and the generation registry follows the last block. Results are a list of records:

```json
{
  "namespace": "global",
  "user": "0",
  "_id": "9631",
  "name": "development_settings_enabled",
  "pkg": "com.android.settings",
  "value": "1",
  "default": "1",
  "defaultSystemSet": "true",
  "history": [
    {
      "timestamp": "2021-11-02 11:21:22.212000",
      "oldValue": "null",
      "newValue": "1",
      "pkg": "com.android.settings"
    },
    {
      "timestamp": "2022-03-14 09:02:11.100000",
      "oldValue": "1",
      "newValue": "0",
      "pkg": "com.example.updater"
    }
  ]
}
```

The change history is the most useful part of the dump that was being discarded: it names the package that changed a setting and when it did so, which for a security-relevant setting is the difference between "accessibility is on" and "accessibility was turned on by `com.example.helper` on this date". `dumpsys` prints those timestamps as `MM-DD HH:MM:SS.mmm` with no year, so they are resolved against the `ending at:` time of the section — the most recent matching date at or before the dump.

## Behaviour delta

- `mvt-android check-bugreport`, `settings` module: results are a list of records instead of a mapping, and the log line counts records.
- Dangerous-setting alerts are unchanged in wording and count, but each now carries the whole record as its event, and takes its event time from the last history entry, so an alert on a changed setting is anchored in time.
- History entries are serialized into the timeline as `settings_change` events (`secure setting "accessibility_enabled" changed from "0" to "1" by com.example.helper`). Settings with no recorded history produce no timeline entries.
- `mvt-android check-androidqf` shares this artifact. Its `settings_*.txt` files are `key=value`, not `dumpsys` records, so it now emits `{namespace, user, name, value}` records rather than a mapping, which keeps a single `check_indicators` for both. Its own `__init__` was pure boilerplate and is gone.

## Related

#912 fixes the metadata-in-value defect for single-line rows by splitting at the first metadata key, and measures the effect across 30 bug reports. This PR reaches the same values by reading whole records, and additionally keeps multi-line values, repeated names, rows without `pkg:`, and the change history. The trailing `tag:`, `isValuePreservedInRestore:` and bare `notPreservedInRestore` tokens that #912 identifies are read as fields of the record in the same way as `default:` and `defaultSystemSet:`, with the bare token recorded as `isValuePreservedInRestore: false`, so a row ending in any of them keeps its plain value whether or not it carries a default. Those rows are the false positives #912 measured.

## Tests

New `tests/android/test_artifact_settings.py` pins multi-line values, the trailing-`default:` leak, the `tag:` and restore-flag tails on rows with and without a default, the generation registry dumped after the last block, repeated names within and across users, a record with no `pkg:`, history resolution across a year boundary, and history with no `ending at:` anchor. `tests/android_bugreport/test_bugreport.py` covers the module end to end including the timeline entry, and the androidqf settings test moves to the new record shape.

Across the Android bug reports available locally, five files carry a settings section (4816 rows; the row counts suggest two distinct devices). None of them prints the vendor tokens, and record counts and dangerous-setting alerts are identical before and after, so the vendor shapes rest on the #912 measurement.

`ruff check`, `mypy` and `pytest` are clean; the six failures in the suite (`tests/common/test_indicators.py` stix2, `tests/test_cli_entry_points.py` plugin completion) reproduce unmodified on `main`.

### DIFF
--- a/src/mvt/android/artifacts/settings.py
+++ b/src/mvt/android/artifacts/settings.py
@@ -4,6 +4,11 @@
 #   https://license.mvt.re/1.1/
 
 import re
+from datetime import datetime
+from typing import Optional, Sequence
+
+from mvt.common.module_types import ModuleAtomicResult, ModuleSerializedResult
+from mvt.common.utils import convert_datetime_to_iso
 
 from .artifact import AndroidArtifact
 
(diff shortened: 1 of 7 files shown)


va-resident:
Ran `_build_record` over the row shapes from the 30 bug reports in #912 —
besides the two tokens you list, `tag:` leaks as well:

```
value:0 tag:null               -> value='0 tag:null'
value:0 notPreservedInRestore  -> value='0 notPreservedInRestore'
```

Both are exactly the false positives I measured in #912 (60 alerts down to 8
across those 30 reports). The value only gets polluted when the row carries no
`default:`, since that's what your `rpartition` anchors on.

Happy to bring the full token list over to this branch, or rebase this onto
#912 — whichever you prefer.

DonnchaC:
Thanks for the feedback and end-to-end testing @va-resident. I've incorporated your fixes into this PR, so it also closes #912.

PR #773, opened 2026-04-12
Show which accessibility services are enabled

besendorf:
## Summary
- Parse both installed and enabled accessibility services from dumpsys output, adding an `enabled` field to each result
- Handle both old (multi-line) and new (AOSP >= 14 single-line) enabled services formats
- Match enabled services against installed services accounting for trailing annotations like `(A11yTool)`
- Log enabled status per service and show enabled count in summary

Fixes #744

## Test plan
- [x] Existing tests updated and passing
- [x] New test for mixed installed/enabled fixture
- [x] Full test suite passes with no regressions
- [ ] Real world test

PR #946, opened 2026-09-22
Carry the accessibility service count the dump states

va-resident:
Fixes #744

Two halves of the same parser, in the order they matter.

**Change**, `artifacts/dumpsys_accessibility.py` and `modules/bugreport/dumpsys_accessibility.py`:

1. **Parse `installedServiceCount=N` per user.** For a user whose services the dump did not list,
   append one record with `component` / `package_name` / `service_name` `None` and
   `installed_service_count=N`, and raise it as **LOW** — a count is a coverage statement, not a
   running service, so it must not compete with one the dump says is enabled. The module's summary
   line reports how many services the dump stated but did not name, so a stated count can no
   longer read as "a total of 0". (You said "at least informational if not higher, could even be
   low" — this is LOW. One line and one assertion if you want it back at informational.)
2. **Name the state in the alert** (`installed, not enabled` / `enabled` / `enabled and bound` /
   `installed, enabled state not stated`) and split its severity, as you asked on this issue: a
   service the dump states is switched **off** is **LOW**, enabled or bound is **MEDIUM**, and a
   dump that does not state the enabled state stays MEDIUM — "not stated" is not "not enabled".
   IOC matching is untouched by the state: `check_app_id()` runs before any of this and still
   returns CRITICAL for a disabled service, which a new test pins.
3. A state section the dump never printed yields `None` instead of `False`, so "not stated" is not
   recorded as "not enabled". `Bound services:{}` (printed, empty) stays `False`.

Both record shapes — a service record and a count-only record — are built from one `_FIELDS` list,
so they cannot drift apart.

**Measured** on 50 bug reports (40 that state a count without listing, 10 that list), upstream vs
patched, running the artifact code directly:

| | upstream | patched |
|---|---|---|
| records | 40 | 80 |
| records with a component | 40 | 40 |
| MEDIUM alerts | 40 | 4 |
| LOW alerts | 0 | 76 |

The 40 new records land one per report on the 40 that produced nothing before. No record is
dropped and no IOC match changes; what moves is the severity of the switched-off services.

Over the wider set of 163 reports the same run gives MEDIUM 305 → 4 and LOW 0 → 407 (301
switched-off services plus 106 stated counts). **The 4 records that stay MEDIUM are the only ones
in the whole set with `enabled=True`** — both belong to the same device, and both are the
accessibility services of an app that is malicious. Every other accessibility record in 163 bug
reports is a service that is installed and switched off.

**Tests:** two new files — a stated count with no listed services produces one record and one LOW
alert, a listed service is unchanged, two users are kept apart, a zero count adds nothing; an
absent section reads `None` while a printed-empty one reads `False`, and the alert text carries
the state.

Two existing assertions change, both in `test_artifact_dumpsys_accessibility.py`: the v14 fixture
never prints an `installed services:` section, so `installed` becomes `None` instead of `False`;
and every service in the Samsung fixture is installed-and-off, so `test_ioc_check`'s three
non-IOC findings are LOW rather than MEDIUM (the IOC match itself is unchanged).

Applies to `main@777db67` with no other change; `pytest tests/android` there goes from 99 passed /
1 skipped to **110 passed / 1 skipped**; `ruff check`, `ruff format --check` and
`mypy --platform linux` (156 files) pass on the branch.

**Related open work** (checked 2026-09-22):

* **#773 "Show which accessibility services are enabled"** (open since 2026-05-04, for this issue)
  is the only other work on this file. Its diff is against the pre-`008844b` parser that no longer
  exists in `main`, it defaults `enabled` to `False`, and it does not touch the count. This PR
  covers the same ground on the current parser; which of the two to take is yours to decide, and I
  have said so on #773 rather than leaving it unmentioned.
* **#839 "Run extraction modules in parallel"** rewrites `modules/bugreport/dumpsys_accessibility.py`'s
  `__init__`; this PR only edits `run()`, so they should not collide.
* **#885 "Add versioned output schemas"** gives every module an `output_model`; the new count-only
  record is a new shape in this artifact's output and would need to be in that schema.


besendorf (changes requested):
I reproduced two issues on commit bdbc07a that should be fixed before merge:

1. **Track state-section presence per user.** In [the normalization loop](https://github.com/mvt-project/mvt/blob/bdbc07a3b32b7f97b24e1c25e97a21c59d09ad98/src/mvt/android/artifacts/dumpsys_accessibility.py#L132-L140), `seen_states` is global. With user 0 listing one installed service but no `enabled services` section and user 10 listing one enabled service, user 0 is incorrectly assigned `enabled=False` and gets a LOW `installed, not enabled` alert. User 10 is also incorrectly assigned `installed=False`. Each user needs its own set of printed sections; please test a multiuser dump with named services and different sections.

2. **Report the unnamed part of a partial count.** [This condition](https://github.com/mvt-project/mvt/blob/bdbc07a3b32b7f97b24e1c25e97a21c59d09ad98/src/mvt/android/artifacts/dumpsys_accessibility.py#L144-L150) suppresses the count-only record once *any* component was named. The existing Android 14 fixture states `installedServiceCount=2` but names only one enabled component. Parsing yields one service record, no count coverage alert, and the summary says one identified with zero more unnamed. Please account for the per-user difference between the stated count and distinct named components, and add a test for that fixture.

The first case incorrectly downgrades an unknown enabled state; the second makes an incomplete service listing look complete.

